Issue #134, opened 2018-01-16
Build instructions and considerations


smortex:
Hi!

I finally managed to build Choria on FreeBSD, the process was unexpectedly complicated, and I am not sure about the exactness of the process I used to build the code.

Because I have no real experience with go-based programs, I looked at other go software packages in the FreeBSD ports tree to see how they are built, and since there is a bunch of differences, it might make sense to consider a few points.

#### How I built choria

In case I did this utterly wrong, let me start by explaining how I built Choria:

```
git clone https://github.com/choria-io/go-choria
cd go-choria
go get
go build
```

This produced a working `go-choria` binary in the working directory \o/.

However, a `~/go` directory was created and a lot of source codes where downloaded there (I assume it's all the choria dependencies… 76MB :fearful:).

#### Reproducible builds

After creating a new user account and building choria the same way (but at a different date), the content of the `~/go` directory is not strictly the same (89 MB this time).  Maybe I am wrong, but my guess is that the dependencies are not fetch against a particular commit, and at some point, the build may break because of a dependency got updated.

In order to package Choria on FreeBSD, the checksum of all sources must be registered so that they are checked before building. Most go applications in the FreeBSD ports tree have all their dependencies included into their source code (!), generaly in a `vendor` directory (e.g. [aptly](https://github.com/smira/aptly/tree/master/vendor), [hub](https://github.com/github/hub/tree/master/vendor), [syncthing](https://github.com/syncthing/syncthing/tree/master/vendor), etc).  While this is arguably ugly, it offers the benefits of reproducibility.  Another option would be to add information about each dependencies version / commit, so that they could be checked.  This is for example done for the [go-cve-dictionary](https://svnweb.freebsd.org/ports/head/security/go-cve-dictionary/Makefile?revision=447954&view=markup#l24) port, based on a file [Gopkg.lock](https://github.com/kotakanbe/go-cve-dictionary/blob/master/Gopkg.lock) in upstream's repository (a lot more work for porters, but far less ugly at the repository level).

Moving into one of these directions (or something similar) would be awesome.

#### Binary name

The repository is called `go-choria`, and `go build` produce a `go-choria` binary. Can you confirm that the file should be renamed to `choria` when installed on the end-user system?

ripienaar:
your build would be:

```
git clone https://github.com/choria-io/go-choria
cd go-choria
go get github.com/Masterminds/glide
glide install
go build .
```

though the actual build command in reality would be a lot more complex, along the lines of:

```
[...]

PR #136, opened 2018-01-16
(#134) Remove duplicate reference to choria-io/go-nats

smortex:
choria-io/nats redirects to choria-io/go-nats, so only use the new name.
The old name's version was more restrictive so move it arround.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -24,8 +24,6 @@ import:
   - context
 - package: github.com/onsi/gomega
   version: ~1.1.0
-- package: github.com/nats-io/nats
-  version: ^1.2.2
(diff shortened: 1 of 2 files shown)


glide.yaml
```diff
@@ -10,7 +10,7 @@ import:
 - package: gopkg.in/alecthomas/kingpin.v2
   version: ^2
 - package: github.com/nats-io/go-nats
-  version: ^1
+  version: ^1.2.2
```
ripienaar: you can make this ^1 again its fine

smortex:
Just amended the commit.  Thanks!

ripienaar:
thanks :)

Issue #243, opened 2018-04-17
create a ruby compatible layer in the RPC system


ripienaar:
At present there is a mcorpc layer that speaks the mcollective RPC protocol and is accessible from the mco cli etc

There's a good chance that mco will not ship in Puppet 6 so we should probably more urgently look at the ability to invoke mco agents from within this new daemon.

We should:

 * Vendor the RPC support code into the go system and it should put it on the OS on demand
 * Have a means of discovering agents that are built using ruby
 * Provide shims in go for each discovered agent that calls out to a ruby helper to do the running

The shims that call ruby should set things up so that authorization and everything works. The outcome should be an entirely mcollective free deploy if possible but there might be some middle ground that's acceptible

To get there we'll probably need a little converter of at least the agent DDLs into JSON along with the ruby DDL so that the go code can figure out what are real actions etc, it's not REALLY needed but will help keep things quicker and simpler.  We must also know the agent timeout which is in the DDL.

Since we now own all the mco plugins that should be easy enough to do

 - [x] Press JSON DDL files during plugin package process (https://github.com/choria-io/puppet-mcollective/pull/167)
 - [x] Interpret JSON DDL files
 - [x] Create a ruby shim that can run an agent given a request as JSON (https://github.com/choria-io/puppet-choria/commit/e01ad7ac33790659bb23b666dbd81eb253805fe6)
 - [x] Create mcorpc agent shim that can present actions for all actions found in a DDL and execute them via the shim
 - [x] Create a concept of agent provider, mcorpc is a agent provider, other types of agents too. Call all providers to load their agents
 - [x] Update choria module to configure the server and shim config (https://github.com/choria-io/puppet-choria/commit/e01ad7ac33790659bb23b666dbd81eb253805fe6)
 - [x] Figure out how to deploy the shim (https://github.com/choria-io/puppet-choria/commit/e01ad7ac33790659bb23b666dbd81eb253805fe6)
 - [x] Install core agents as a provider, move under mcorpc
 - [x] ~Move additional agents handling into mcorpc core provider~ not a mco rpc feature its generic
 - [x] ~Extract mcorpc provider into own project~ defer to https://github.com/choria-io/go-choria/issues/251 once stable

ripienaar:
Proposed JSON DDL in https://github.com/choria-io/puppet-mcollective/pull/167

ripienaar:
Basic shimm to call an agent:

 - [x] same timeout handling as mcollectived
[...]

PR #244, opened 2018-04-17
(#243) Interpret JSON DDL files

ripienaar:
This creates a fairly light interpreter for the JSON representations
of DDL files

Main aim is to get access to valid action lists and timeout but some
other bits are there too


ripienaar:
WIP dont merge

PR #268, opened 2018-04-25
(#267) correctly pass request time to ruby shim

ripienaar:
The request time was incorrectly passed to the ruby shim resulting
in null timestamps in logs and failure in anything that needed the
real timestamp


ripienaar:
build passes, merging on red because circleci is broken

Issue #271, opened 2018-04-25
rotate choria-*.log


ripienaar:
Previously in https://github.com/choria-io/go-choria/issues/269 we made a specific rotation for the audit log but i think `choria-*.log` is a good bet, this way things like agent frameworks - like mcorpc - can log into `choria-mcorpc.log` for example

PR #272, opened 2018-04-25
(#271) Rotate choria-*.log

ripienaar:
To allow things like the mcollective compat framework to log and
be rotated we create a pattern where they will all go


packager/templates/debian/global/choria-logrotate
```diff
@@ -1,4 +1,4 @@
-/var/log/{{cpkg_name}}-audit.log
```
ripienaar: @smortex if you do do logrotation setup this is going to be the safest bet as things like the old mco log will fit into this pattern

PR #280, opened 2018-04-30
(misc) attempt to build a msi

ripienaar:


ripienaar:
leaving this hell for some other time

Issue #353, opened 2018-06-25
improve STAN adapter connection error handling


ripienaar:
https://github.com/nats-io/go-nats-streaming#connection-status

PR #354, opened 2018-06-26
(#353) handle STAN connection errors

ripienaar:
This uses the NATS Streaming feature found in streaming server >= 0.10.0
to detect and handle stream ping timeouts


ripienaar:
WIP untested

Issue #362, opened 2018-07-08
maintain an agent disable list


ripienaar:
Agent providers should have a global list of what agents they are not allowed to start, in backplane for example there is very little reason for rpcutil and choria_util yet one can't turn them off

PR #371, opened 2018-07-11
(#362) avoid nil pointer

ripienaar:


ripienaar:
merged on transient unrelated build error

Issue #444, opened 2018-09-11
socket timeout errors due to max. open files limit on broker


mpepping:
Conservative limitations of the 'Max. open files' setting on some (enterprise) Linux distributions causes the Choria Broker to stop handle new connections when the Max open files limit is reached. When this condition is hit, the Choria Broker logs the following for any subsequent connection:

```
{"component":"network_broker","level":"error","msg":"Temporary Client Accept Error (accept tcp [::]:4222A: accept4: too many open files), sleeping 10ms","time":"2018-09-06T08:07:23Z"}
```

On RHEL7 (for example) the 'Max. open files' limit defaults to 1.024. Adding over ~1.000 nodes to the Choria Broker on RHEL7 cause a `NATS::IO::SocketTimeoutError` message on the nodes that are unable to connect to the broker. 

Setting the systemd `LimitNOFILE` parameter with an increased value (>1.024) in the Choria Broker systemd unit-file resolves the issue. Since Choria aims for max. 50.000 nodes per broker. It would be useful to set sane default `LimitNOFILE` values in the systemd unit-files for the broker (e.g., `LimitNOFILE=51200`).

PR #445, opened 2018-09-11
(#444) Sets a default LimitNOFILE=51200 for Choria Broker.

mpepping:
Since a default Choria deployment aims for (max.) 50.000 nodes, it is useful to set a matching value for the max. open files in the systemd unit-file. 

Tested on RHEL7 and Debian Jessie.

### DIFF
--- a/packager/templates/debian/generic/broker.service
+++ b/packager/templates/debian/generic/broker.service
@@ -5,6 +5,7 @@ After=network.target
 [Service]
 StandardOutput=syslog
 StandardError=syslog
+LimitNOFILE=51200
 User=root
 Group=root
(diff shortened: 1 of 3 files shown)


mpepping:
@ripienaar Good suggestion. I chose to add the setting to the init-script, so that users are able to overwrite this settings value via `/etc/sysconfig/choria-broker`, since manual changes to the init-script file can be lost on package upgrades. The same method applies to the systemd units, where users can set a custom LimitNOFILE value via a systemd drop-in unit.

Please let me know if it is worth to extend the Choria documentation on this.

Tested on EL6.

ripienaar:
very good thank you so much

Issue #470, opened 2018-10-27
replace github.com/satori/go.uuid with github.com/gofrs/uuid


ripienaar:
The satori version has weak random integration and can come up with some weak series of uuid's as per https://github.com/satori/go.uuid/issues/73

Probably not a big deal for this project but nonetheless moving is best as per https://github.com/satori/go.uuid/issues/84

PR #471, opened 2018-10-27
(#470) replace satori/go.uuid with gofrs/uuid

ripienaar:
The satory/go.uuid is abandoned and uses rand in an unsafe way
while the gofrs one is maintained but has an unfortunate API change
to protect against unsafe UUID generation

Rock and a hard place but ultimately making this breaking API change
is the right things to do


ripienaar:
compile error here is expected until dependencies are updated - they cant update till this is update, doh

Issue #508, opened 2018-12-19
Set environment variables in Rakefile for MCOLLECTIVE_CERTNAME


vjanelle:
Not everyone has a username of `rip`.  This makes it harder to run `rake test` without pre-setting up environment variables.

Set the default expected `MCOLLECTIVE_USERNAME` in the Rakefile for ease of development.

ripienaar:
thanks - aside this is a minor thing you can do this kind of thing without a ticket

PR #509, opened 2018-12-19
(#508) We are not all named `rip`

vjanelle:
Sets the default `MCOLLECTIVE_CERTNAME` in the Rakefile to make this
happier for other developers running tests outside the docker container
(ex, MacOS X)


ripienaar:
this is great not sure why I never did this - probably bacause I often just run ginkgo -r

PR #538, opened 2019-01-13
Run broker/server as a user if specified on el6

vjanelle:
This is more useful for the actual broker than the server, but will
honor specified user value from sysconfig override.

Note: Changing the value of the user without changing the value of the
pid file or logfile will probably result in a bad time.  There is also
no good way of changing these values unless the broker or server are
shut down first, since the init script depends on the pid file.

### DIFF
--- a/packager/templates/el/el6/broker.init
+++ b/packager/templates/el/el6/broker.init
@@ -22,6 +22,9 @@ lockfile="/var/lock/subsys/${prog}"
 logfile="/var/log/${prog}"
 conffile="{{cpkg_etcdir}}/broker.conf"
 
+# If you're going to change the value of $user for non-root operations, make sure you change the value of
+# $pidfile and $logfile as well, otherwise you won't be able to start the daemon correctly.
+
 # set the open file limit to allow over 1024 connections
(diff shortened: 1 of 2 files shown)


ripienaar:
Wondering if it’s worth the complexity to add this. 

You don’t have EL7? Easier with systemd to override things. Otherwise how would you feel about me releasing a broker container? I want to do that anyway because I want to support deploying with helm

vjanelle:
I’m not entirely on centos 7 yet.  And centos 6 with docker isn’t really useful, IMHO.  

I’m ok with keeping this as a private change, but the complexity is nothing new with sysvinit scripts.

ripienaar:
How do you intend to actually change this setting? you do your own packages? As is there isnt really anyway for a choria user to actually affect this change, so if we want this a toggle we should add it as a variable users can set in the buildsettings.yaml for example

vjanelle:
@ripienaar I create /etc/sysconfig/choria-broker via puppet.  But, right now have my own forked modules.  If the value exists in there, it'll get overridden, but defaults to root in the init script.

PR #651, opened 2019-07-30
first draft of rpc go client

snakeb1t:
JSON is my preferred output, so that's the only mode right now, though I'm not averse to adding more output modes.

Feel free to suggest changes!

### DIFF
--- a/cmd/rpc.go
+++ b/cmd/rpc.go
@@ -1,0 +1,337 @@
+package cmd
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"github.com/choria-io/go-choria/choria"
+	"github.com/choria-io/go-client/discovery/broadcast"
+	"github.com/choria-io/go-protocol/protocol"
+	"github.com/choria-io/mcorpc-agent-provider/mcorpc/client"
+	ddl "github.com/choria-io/mcorpc-agent-provider/mcorpc/ddl/agent"
+	"os"
+	"regexp"
+	"strings"
+	"sync"
+	"time"
+)
+
+var libDirs = []string{
+	"/opt/puppetlabs/mcollective/plugins",
+}
+var allFilters = protocol.NewFilter()
+
+type jsonDisplay struct {
+	client.RPCReply
+
+	Agent  string `json:"agent"`
+	Action string `json:"action"`
+	Sender string `json:"sender"`
+}
+
+type rpcCommand struct {
+	command
+
+	noResults bool
+
+	one        bool
+	batch      int
+	batchSleep int
+	limitSeed  string
+	limit      string
+	json       bool
+	display    string
+	config     string
+	verbose    bool
+
+	target            string
+	discoveryTimeout  int
+	timeout           int
+	quiet             bool
+	ttl               int
+	replyTo           string
+	discMethod        string
+	discOptions       []string
+	nodes             string
+	publishTimeout    int
+	threaded          bool
+	sort              bool
+	connectionTimeout int
+
+	with         []string
+	sel          []string
+	withFact     []string
+	withClass    []string
+	withAgent    []string
+	withIdentity []string
+
+	agent  string
+	action string
+	args   map[string]string
+}
+
+func (r *rpcCommand) Setup() (err error) {
+	r.args = make(map[string]string)
+	r.cmd = cli.app.Command("rpc", "Generic RPC agent client application")
+
+	// TODO: implement me
+	//r.cmd.Flag("no-results", "Do not process results, just send request").BoolVar(&r.noResults)
+
+	// TODO: implement me
+	//r.cmd.Flag("one", "Send request to only one discovered nodes").Short('1').BoolVar(&r.one)
+	r.cmd.Flag("batch", "Do requests in batches").PlaceHolder("SIZE").IntVar(&r.batch)
+	r.cmd.Flag("batch-sleep", "Sleep time between batches").IntVar(&r.batchSleep)
+	// TODO: implement me
+	//r.cmd.Flag("limit-seed", "Seed value for deterministic random batching").StringVar(&r.limitSeed)
+	// TODO: implement me
+	//r.cmd.Flag("limit", "Send request to only a subset of nodes, can be a percentage").StringVar(&r.limit)
+	// json is default and only means of output at this time (and perhaps forever)
+	//r.cmd.Flag("json", "Produce JSON output").Short('j').BoolVar(&r.json)
+	r.cmd.Flag("display", "Influence how results are displayed. One of ok, all or failed").StringVar(&r.display)
+	r.cmd.Flag("verbose", "Be verbose").Short('v').BoolVar(&r.verbose)
+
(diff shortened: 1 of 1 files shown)


cmd/rpc.go
```diff
@@ -0,0 +1,337 @@
...
+	ddl "github.com/choria-io/mcorpc-agent-provider/mcorpc/ddl/agent"
+	"os"
+	"regexp"
+	"strings"
+	"sync"
+	"time"
+)
+
+var libDirs = []string{
+	"/opt/puppetlabs/mcollective/plugins",
```
ripienaar: you can grab this from `cfg.LibDir`

cmd/rpc.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+func (r *rpcCommand) Configure() error {
+
+	err = commonConfigure()
+	if err != nil {
+		return err
+	}
+
+	if r.threaded {
```
ripienaar: I think a lot of these if statements here can be simplified by either having the cli options set the options directly and by letting the cli parsing default to whats in the config

ripienaar: I do something like this which simplify filter stuff a lot:

```
var (
	factF     []string
	agentsF   []string
	classF    []string
	identityF []string
	combinedF []string
)

func addRPCFilter(cmd *kingpin.CmdClause) {
	cmd.Flag("wf", "Match hosts with a certain fact").Short('F').StringsVar(&factF)
	cmd.Flag("wc", "Match hosts with a certain configuration management class").Short('C').StringsVar(&classF)
	cmd.Flag("wa", "Match hosts with a certain Choria agent").Short('A').StringsVar(&agentsF)
	cmd.Flag("wi", "Match hosts with a certain Choria identity").Short('I').StringsVar(&identityF)
	cmd.Flag("with", "Combined classes and facts filter").Short('W').PlaceHolder("FILTER").StringsVar(&combinedF)
	cmd.Flag("batch", "Do requests in batches").Default("0").IntVar(&batchSize)
	cmd.Flag("batch-sleep", "Sleep time between batches").PlaceHolder("SECONDS").Default("10").IntVar(&batchSleep)
	cmd.Flag("nodes", "Text file with nodes to use for discovered noded").PlaceHolder("FILE").ExistingFileVar(&nodesFile)
}

func rpcFilter(agent string) (*protocol.Filter, error) {
	return client.NewFilter(
		client.FactFilter(factF...),
		client.AgentFilter(agentsF...),
		client.ClassFilter(classF...),
		client.IdentityFilter(identityF...),
		client.CombinedFilter(combinedF...),
	)
}
```

snakeb1t: I get segfaults when trying to simplify the if statements. For example:

```
r.cmd.Flag("target", "Target messages to a specific sub collective").Short('T').Default(cfg.MainCollective).StringVar(&cfg.MainCollective)
```

gets me this error when I run it:

```
[vagrant@puppet ~]$ choria rpc -q process list pattern=ruby user=root
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0xba16aa]

goroutine 1 [running]:
github.com/choria-io/go-choria/cmd.(*rpcCommand).Setup(0xc000001c80, 0x0, 0x0)
	/go/src/github.com/choria-io/go-choria/cmd/rpc.go:84 +0x77a
github.com/choria-io/go-choria/cmd.ParseCLI(0xc000155f78, 0xbc14dd)
	/go/src/github.com/choria-io/go-choria/cmd/cmd.go:53 +0x533
main.main()
	/go/src/github.com/choria-io/go-choria/main.go:15 +0x26
```

ripienaar: ah i bet cfg isnt set yet when the cli arguments are parsed, lame :(

snakeb1t: Also, I can't seem to find the FactFilter, AgentFilter, ClassFilter, etc functions, though I was looking in https://github.com/choria-io/go-protocol/blob/master/protocol/filter.go

ripienaar: you generally wouldnt need to muck about with protocol at all, there are 2 types of client.

https://github.com/choria-io/go-client and the rpc one you already know.  It's a bit like a TCP packet vs a HTTP packet, RPC requests are wrapped within normal client requests (they just move bytes).

Like in TCP addressing is in the underlying client, so the go-client have various helpers and discovery stuff etc. So in the example `client.NewFilter()` comes from this lower level client at https://github.com/choria-io/go-client/blob/master/client/filter.go

cmd/rpc.go
snakeb1t: fixed

cmd/rpc.go
snakeb1t: Ok, thanks! I've simplified the filter configuration. Should I put back those if statements in `Configure` because of the segfaults?

ripienaar: yeah i guess, I'll take a look at how i can make it easier when i look again after my trip

snakeb1t: corrections pushed!

ripienaar:
So, been thinking about this, I really want to head in the direction of https://github.com/choria-io/mcollective-choria/issues/72 eventually

So I think I have a few changes I'd like to make here to at least not paint us into a corner:

 * lets call this `choria request` and not `choria rpc`
 * lets add a `-j` setting which will enable JSON mode, you can default it to on for now or print a warning to STDOUT before the JSON if its not set warning people about future change and using `-j` to supress the warning.

Later when I start working on a new cli experience at least we dont have a implied JSON output we have many programs relying on, I can then keep the same behavior you have here and be sure all people who actually rely programataically on JSON output would have used `-j`

what do you think?

snakeb1t:
makes sense to me!

ripienaar:
@snakeb1t this should help with the limit CLI options https://github.com/choria-io/mcorpc-agent-provider/pull/69

cmd/rpc.go
ripienaar: Regarding the segfaults when trying to simplify Configured() - you can do the `commonConfigure()` stuff in `Setup` prior to creating all the app stuff then adjust your `Configure()` to not call `commonConfigured()` again, that should solve it and simplify things a lot

cmd/rpc.go
```diff
@@ -0,0 +1,287 @@
...
+		display := &jsonDisplay{}
+		err := json.Unmarshal([]byte(reply.Message()), display)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to unmarshal incoming message from %s", reply.SenderID())
+			return
+		}
+		display.Agent = r.agent
+		display.Action = r.action
+		display.Sender = reply.SenderID()
+		replies <- *display
```
ripienaar: this could cause trouble, if someone has 2 nodes with the same identity or if some new nodes appeared between discovery and request and a broadcast method is done you will get replied here that you do not expect and so overflow the channel leading to deadlock.

I'd set up a background routine to consume the replies channel and do what needs doing in parallel so it wont fill up.

snakeb1t:
Closing in favor of https://github.com/choria-io/go-choria/pull/662

PR #662, opened 2019-09-04
(#663) add basic client code

ripienaar:


ripienaar:
@snakeb1t  hope you dont mind but I pulled this together from various pieces of closed source code I wrote over the last while where I spent some time to make things behave the way mco rpc behaves today.

LOTS missing here, all the fancy display stuff, but this I am quite happy to ship as a basic client and I added the `-j` too with an improved JSON output that includes stats about the request.

What do you think?

snakeb1t:
👍 !!!

snakeb1t:
Actually, there is one thing I would need: `discovery-timeout`: https://github.com/snakeb1t/go-choria/blob/506db65edb5e2bc328fed7710bb51556d3abefe0/cmd/rpc.go#L90

The default of two seconds is too short when using pkcs11, since the yubikeys we use need a key press to verify a crypto operation is legit.

snakeb1t:
Would be cool to load a client config and get the discovery timeout from that too, so we don't have to provide it for each invocation

ripienaar:
OK - config is already loaded, just need to handle that override then I guess.

Issue #703, opened 2019-10-14
Add option to output result to file


snakeb1t:
Handy in general, but especially useful with pkcs11 security module, as the pin prompt outputs to stdout.

ripienaar:
thanks, doh, I forgot to squash it, anyway thanks :)

PR #704, opened 2019-10-14
(#703) Add output option

snakeb1t:
Tested with/without -j and -v

### DIFF
--- a/cmd/req.go
+++ b/cmd/req.go
@@ -56,8 +56,8 @@ type reqCommand struct {
 	combinedF        []string
 	outputFile       string
 
-	w  *bufio.Writer
-	fh *os.File
+	outputWriter     *bufio.Writer
+	outputFileHandle *os.File
 }
 
 type rpcStats struct {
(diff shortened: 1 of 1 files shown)


cmd/req.go
```diff
@@ -110,6 +114,7 @@ func (r *reqCommand) Setup() (err error) {
 	r.cmd.Flag("verbose", "Enable verbose output").Short('v').BoolVar(&r.verbose)
 	r.cmd.Flag("display", "Display only a subset of results (ok, failed, all, none)").EnumVar(&r.displayOverride, "ok", "failed", "all", "none")
 	r.cmd.Flag("discovery-timeout", "Timeout for doing discovery").PlaceHolder("SECONDS").IntVar(&r.discoveryTimeout)
+	r.cmd.Flag("output-file", "Filename to write output to").PlaceHolder("FILENAME").Short('o').StringVar(&r.outputFile)
```
ripienaar: you can use `FileVar` or `OpenFileVar` and it will open it for you if supplied etc, granted its not a HUGE improvement but a bit I guess.

cmd/req.go
```diff
@@ -54,6 +54,10 @@ type reqCommand struct {
 	classF           []string
 	identityF        []string
 	combinedF        []string
+	outputFile       string
+
+	w  *bufio.Writer
```
ripienaar: mind giving these 2 more descriptive names?

cmd/req.go
```diff
@@ -471,6 +486,8 @@ func (r *reqCommand) displayResultsAsTXT(res *rpcResults) error {
 	nodeListPrinter(res.Stats.NoResponses, "No Responses from")
 	nodeListPrinter(res.Stats.UnexpectedResponses, "Unexpected Responses from")
 
+	r.w.Flush()
```
ripienaar: We should flush in the loop that prints all the results after each result else the UI will pause weirdly till the buffers are full.

ripienaar: So end of ddlAssistedPrinter, basicPrinter and errorPrinter should all flush at their end

snakeb1t: Done

cmd/req.go
snakeb1t: Done

cmd/req.go
snakeb1t: I still need to check if the var was used, so I just elected to keep as is, if that's ok

Issue #778, opened 2020-01-23
Ensure that Windows installer creates folder under `Program Files` instead of `Program Files (x86)`


mrbanzai:
In our research and testing, it seems that you have to set a couple of additional attributes in order for the generated Windows installer to create the program directory in the proper location.  PR forthcoming.

ripienaar:
Related to https://github.com/choria-io/puppet-choria/issues/145

thanks!

PR #782, opened 2020-01-23
(#778) force architecture in wixl command

mrbanzai:
Overlooked that we had been forcing the architecture to the `wixl` command via `-a x64` in our testing.  This updates #778, and has been tested to generate a functional MSI using the build process.


ripienaar:
thanks

PR #829, opened 2020-02-17
(#831) attempt to cache and reused transports for batches

ripienaar:


providers/agent/mcorpc/client/options.go
```diff
@@ -148,6 +148,21 @@ func (o *RequestOptions) ConfigureMessage(msg *choria.Message) (err error) {
...
+	//
+	// (TTL + DiscoveryTimeout + Timeout) * batches
+	//
+	// We have to allow TTL per batch since the last batch will get it much
+	if msg.IsCachedTransport() && o.BatchSize != len(o.Targets) {
+		batches := int(math.Ceil(float64(len(o.Targets)) / float64(o.BatchSize)))
+
+		msg.TTL = batches * (msg.TTL + int(o.DiscoveryTimeout.Seconds()) + int(o.Timeout.Seconds()))
+		if msg.TTL > int((24 * time.Hour).Seconds()) {
+			return fmt.Errorf("cached transport TTL is unreasonably long")
```
ripienaar: @snakeb1t I added this check here, the reaon is that the TTL is the window of packet replay - someone capturing your signed packet can replay it within TTL so generally you want it short but you definitely do not want it very long.

Will this bite you do you think?

snakeb1t: I think it will because I have a large cluster and a use case where I need small batches to run. Can we clamp to 24 hours? Or better yet, make it a config value?

ripienaar: Yeah so as is it’s a day. Good yeah? Rather add a config option when we know we need it

snakeb1t: Yeah, having a max of one day is good enough for me. I don't expect batches to run longer than that

snakeb1t:
To be clear, the error should be replaced with clamping down the value to 24h

ripienaar:
Hmm if I will cap it and change it without warning I will be much more aggressive like 5 hours max which should be plenty for what choria is used for. The risk of high TTLs and request replay is quite big

snakeb1t:
Sounds good. Implementing a replay cache can mitigate too

ripienaar:
An efficient cache is effectively impossible to build, its either in memory on individual servers or spooled to disk - regardless choria being loosyly discovered and loosely assembled means at any point new machines can come up and would have no cache - and so can be replayed against.  There's no central service or central DB - and having such a thing would basically break major assumptions

snakeb1t:
Hey @ripienaar, I was hoping that instead of erroring if TTL is greater than five hours, we just set it to 5 hours and continue. Does that sound ok?

ripienaar:
It doesn’t really make sense. At the end of TTL as the code stands it won’t make a new message allowing you to press the 2FA

We could detect we are near end of TTL and three invalidate that message cache but that’s not how it works now. 

Do you really have invocations where a single RPC takes 5 hours?

snakeb1t:
I don't, but the calculation doesn't know that, since it's looking at the agent timeout, which can be much higher than the runtime

ripienaar:
Any invocation can only ever be max agent time + network overhead because agents will timeout anyway then and get killed off. 

So multiply that by batches and you get your total possible runtime - any longer and the agents would have been killed. 

So batches*longest possible single run is exactly right isn’t it? So if you don’t have more than 5 hour runtimes you are fine.

ripienaar:
As for "much higher" TTLs, the shell and task agents are built around short actions with long backgrounded tasks, that's the right approach, much of mcollective (less in choria but still valid) assumes short lives action, single digit minutes are good numbers.

PR #835, opened 2020-03-03
(#834) Add buster_64 support

vjanelle:


vjanelle:
@ripienaar this doesn't test building the packages I take it?

ripienaar:
Also need to be added here https://github.com/choria-io/go-choria/blob/master/.circleci/config.yml

At present nothing tests building of packages, we need that but dont have it now - only packages being built is the nightly RPMs

.circleci/config.yml
```diff
@@ -143,6 +143,19 @@ jobs:
...
+      - image: circleci/golang:1.14
+
+    working_directory: /go/src/github.com/choria-io/go-choria
+
+    environment:
+      PACKAGE: buster_64
+      BUILD: foss
+      BUILDER: choria/packager:buster-go1.14
+
+    steps: *std_build_steps
```
ripienaar: also down around line 252

Issue #687, opened 2019-09-16
req subcommand: add option to require specifying a filter


snakeb1t:
This avoids the issue where a user mistakenly sends a command to all hosts, rather than a filtered set of hosts.

ripienaar:
Maybe better as a config option so all clients are sibject to it? And we add this feature to the rpc client lib?

snakeb1t:
Yeah, I added a config option here: https://github.com/choria-io/go-config/issues/27

Didn't think about adding the feature to the rpc client lib, though. I was thinking this was more for human use than automation, which is why I was thinking of just adding it to `req` subcommand

snakeb1t:
So if we implemented it in rpc client lib, it would be an option to `broadcast.Discover`? If that were the case, we wouldn't want the default to be true, so you would have to modify all current clients anyways, to pass the dopt in

ripienaar:
Probably rather the mcorpc client, I am super weary of over engineering the core client.

PR #838, opened 2020-03-13
(#687) support disabling requests without filters

ripienaar:
This moves the config item to plugin to avoid breaking the ruby
config parser that fails on unknown things

Additionally improves handling of discovery messages when detecting
if a message is unfiltered


snakeb1t:
Thanks for setting up the unfiltered check!

PR #853, opened 2020-04-07
(#852) Install choria binary in /usr/bin

### LINKED ISSUE #852: install choria in the bin, not the sbin directory
The choria package currently installs `choria(1)` as `/usr/sbin/choria`.

This initially made sense because the choria binary as only a server component.  But recently, choria gained client capabilities (e.g. `choria req`) and regular users usually do not have `/usr/sbin` directory in their `$PATH`, making the `choria` harder to find by new users.

The packaging of choria should rather install choria as `/usr/bin/choria`.

smortex:
/usr/sbin was previously chosen because choria only provided services
components.  Now that choria also has client capabilities, install it to
/usr/bin which is in the $PATH of all users by default.

Fixes #852

### DIFF
--- a/packager/buildspec.yaml
+++ b/packager/buildspec.yaml
@@ -50,7 +50,7 @@ foss:
     defaults:
       name: choria
       display_name: Choria
-      bindir: /usr/sbin
+      bindir: /usr/bin
       etcdir: /etc/choria
       release: 1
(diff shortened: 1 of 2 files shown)


ripienaar:
sweet, easy enough :) thanks!

PR #908, opened 2020-06-24
Build and ship ARM packages

smortex:


.circleci/config.yml
```diff
@@ -216,6 +229,13 @@ jobs:
               package_cloud push choria/${REPO}/el/${d} ${i}
             done
           done
+          for d in buster
+          do
+            for i in $(find /tmp/workspace -name \*.${d}.\*.deb)
```
smortex, line 234: I could not find what I an supposed to be searching here, so maybe this need adjustment…

.circleci/config.yml
```diff
@@ -216,6 +229,13 @@ jobs:
...
+            do
+              package_cloud push choria/${REPO}/debian/${d} ${i}
```
smortex, line 236: This path might also need adjustment.

.circleci/config.yml
ripienaar: yeah it would be the same as rake build, I tend to limit what I stick on nightly because the repo grows quite a lot quite quickly so maybe try to limit the find to the arm ones for now

.circleci/config.yml
ripienaar: yeah we'll see - these CI definitions are a pain, ok by me if they fail a night or two while we get it right. 

Anticipating issues maybe stick a `find /tmp/workspace` in there just to help us debug so we know whats there

.circleci/config.yml
smortex: Oh, I see, I didn't realise there was no Debian nightly releases, and the problem does not exist with CentOS / RedHat.

smortex:
I adjusted to PR according to your feedback, if everything is good with the nightly packages, we would just have to add a bit of configuration for building release packages, and maybe drop the nightly packages (I do not think I would need them in the long term).

Thanks!

ripienaar:
OK sounds good, I'll be doing a release early next week probably

ripienaar:
lets see ho it goes

Issue #907, opened 2020-06-24
Provide ARM packages


smortex:
ARM packages could be handy :tada:

ripienaar:
@smortex how we're doing here? happy with the outcome? I am looking to do a release in the next day or 2, we should wire things up for release

smortex:
Yes!  Amazing :heart_eyes:!  Let's adjust this to distribute releases and stop daily distribution of nightlies (unless it's useful to someone else?)  Stay tuned!

ripienaar:
yeah we can drop the nightlies, did them just for your testing

ripienaar:
thank you :)

PR #916, opened 2020-06-30
(#907) Distribute release Debian Buster ARM packages

smortex:
The nightly armhf packages are working as expected, switch to releasing packages for choria releases and stop building nightlies.


ripienaar:
👍

Issue #987, opened 2020-08-25
Improve AAA model


ripienaar:
We'd like to support connecting a leafnode to an existing choria network without decreasing the security posture of that network but letting the leafnode be clients only and on TLS that doesnt require people to issue certs etc. We want to ensure these leafnodes thats configured in this way can only be used by clients not servers.

```
client -- unverified tls -> leafnode - verfied tls -> choria cluster
       \- unverified tls -> AAA server
```

So to get this we need a few things:

Clients:

 * Only when AAA is configured in the client, support disabling tls veriication in the config file `plugin.security.client_anon_tls`

Leafnode:

 * Support unverfied tls on the client port only when leafnodes are defined - `plugin.choria.network.client_anon_tls`  
 * Support for configuring client-only ACLs via a config flag `plugin.choria.network.deny_server_connections` and setting `plugin.choria.network.client_hosts` enable this

AAA Server

 * Support unverified TLS on the listen port - think it already does

/cc @snakeb1t

ripienaar:
With above PR this config does broker and client.

First we assume the already existing choria network is running with `plugin.choria.network.leafnode_port = xxxx` to enable listening for leafnodes - this has to be a different port from client port.

Now we can set up a new nats broker thats:

 * A leafnode
 * Disables TLS verification
 * Connects to choria with a trusted cert and use same to listen locally but ultimately this will not be used as verify is off, it just needs something

```
loglevel = debug

# enables the broker
plugin.choria.broker_network = true

# listens for clients
plugin.choria.network.client_port = 4222
plugin.choria.network.peer_port = 0

# enables leafnodes and sets up a remote using choria certs
plugin.choria.network.leafnode_port = 9222
plugin.choria.network.leafnode_remotes = choria
plugin.choria.network.leafnode_remote.choria.url = nats://choria.example.net:7422
plugin.choria.network.leafnode_remote.choria.tls.cert = /home/rip/.puppetlabs/etc/puppet/ssl/certs/leafnode.pem
plugin.choria.network.leafnode_remote.choria.tls.key = /home/rip/.puppetlabs/etc/puppet/ssl/private_keys/leafnode.pem
plugin.choria.network.leafnode_remote.choria.tls.ca = /home/rip/.puppetlabs/etc/puppet/ssl/certs/ca.pem

# sets anon tls, only possible when a leafnode is defined
plugin.choria.network.client_anon_tls = true

# disable server connections
plugin.choria.network.deny_server_connections = true

[...]

PR #988, opened 2020-08-25
(#987) initial support for leafnode and non verified TLS clients

ripienaar:

### DIFF
--- a/choria/connection.go
+++ b/choria/connection.go
@@ -2,6 +2,7 @@ package choria
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"net/url"
 	"strings"
@@ -14,6 +15,8 @@ import (
 	"github.com/choria-io/go-choria/config"
 	"github.com/choria-io/go-choria/protocol"
 	"github.com/choria-io/go-choria/srvcache"
+	"github.com/choria-io/go-choria/tlssetup"
+
 	nats "github.com/nats-io/nats.go"
 	log "github.com/sirupsen/logrus"
 )
@@ -583,7 +586,22 @@ func (conn *Connection) Connect(ctx context.Context) (err error) {
 		}),
 	}
 
-	if !(conn.config.DisableTLS || conn.choria.ShouldUseNGS()) {
+	switch {
+	case conn.config.Choria.ClientAnonTLS && !conn.config.InitiatedByServer:
+		conn.logger.Debug("Setting anonymous TLS for NATS connection")
+
+		cfg := tlssetup.TLSConfig(conn.config)
+		tlsc := &tls.Config{
+			MinVersion:               tls.VersionTLS12,
+			PreferServerCipherSuites: true,
+			CipherSuites:             cfg.CipherSuites,
+			CurvePreferences:         cfg.CurvePreferences,
+			InsecureSkipVerify:       true,
+		}
+
+		options = append(options, nats.Secure(tlsc))
+
+	case !(conn.config.DisableTLS || conn.choria.ShouldUseNGS()):
 		tlsc, err := conn.choria.TLSConfig()
 		if err != nil {
 			err = fmt.Errorf("could not create TLS Config: %s", err)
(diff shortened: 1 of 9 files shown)


ripienaar:
@vjanelle asked for your review just for initial sanity checking atm, I'll add some tests etc before merging later

choria/connection.go
```diff
@@ -583,15 +586,31 @@ func (conn *Connection) Connect(ctx context.Context) (err error) {
...
+	case conn.config.Choria.ClientAnonTLS && !conn.config.InitiatedByServer:
+		conn.logger.Debug("Setting anonymous TLS for NATS connection")
+
+		cfg := tlssetup.TLSConfig(conn.config)
+		tlsc := &tls.Config{
+			MinVersion:               tls.VersionTLS12,
+			PreferServerCipherSuites: true,
+			CipherSuites:             cfg.CipherSuites,
+			CurvePreferences:         cfg.CurvePreferences,
+			InsecureSkipVerify:       true,
```
vjanelle, line 599: The intention is to accept any client cert?

ripienaar: Yes, this would be a extra choria broker whose purpose is specifically to accept clients using the AAA server - who don’t have certs at all. 

This lets them connect without lowering the security of the existing choria network. 

Combined with the setting to disable servers and limiting clients from known locations this will be a safe way for clients only. I might require that those 2 settings are configured before enabling this mode

More background in slack and the issue.

ripienaar:
@snakeb1t Do you have a way to test this PR for your needs? I wrote up a little guide here https://gist.github.com/ripienaar/5d2477b10f7fafc759acffa13c69dc6e

snakeb1t:
The only way I can test this end to end is to implement the login subcommand and patch the req subcommand, i think

ripienaar:
`req` should work already as is, and I can show you do to do mco login using curl maybe would that help?

ripienaar:
```
curl -s --request POST -d '{"username":"puppetadmin", "password":"secret"}' -H "Content-type: application/json" -k https://localhost:8080/choria/v1/login | jq -r .token > .choria.token
```

That's all that login does

snakeb1t:
Oh wow, that makes sense! I should be able to test that sometime. For the final iteration, I would want some pkcs11 goodness with the login command

ripienaar:
What pkcs11 features?

snakeb1t:
The ability for the login command to perform an x509 authentication with an identity provider using a pkcs11 device

ripienaar:
Hmm that might well be a seperate login protocol then though? Since the AAA service login only speaks user/password at present, I am not sure how that would gel.

I can see how AAA service could log into the IDP on behalf of a user given a username/pasword as above, but if you want to use a pkcs11 device what would you send to AAA service for login creds to speak to keycloak? 

Am still not up to scratch with pkcs11 so excuse the dumb questions

snakeb1t:
I think what you do is: login command looks for jwt, can't find it, sends AAA service a generic message saying, i wanna auth. AAA service then redirects to idp, and somehow the idp requests x509 auth, which is where the login command takes over and does the pkcs11 dance. The idp x509 request part i'm hazy on, so I'll do some research

ripienaar:
Hmm, OK, afraid that would have to be a different `login` command then since effectively you wouldnt use AAA service login url then (have no redirect support etc)

That's fine this is supported AAA service lets you split the 2 roles and even just run one.

Maybe what we should do is let `choria login` check if you have, for eg, `/bin/choria-login` command then just run that.  This way you can build whatever you like as long as we end up with a JWT in a file its all good?

snakeb1t:
What if I expose a new endpoint on AAA service that does redirects to configured idp address if no jwt is provided in request? That would follow the saml-like workflows I'm used to

ripienaar:
I suppose we can add a extra field here that would indicate redirect

https://github.com/choria-io/aaasvc/blob/master/api/gen/models/login_response.go

snakeb1t:
After reading up on OIDC flows, it seems that what I was envisioning was close to the "implicit flow", so instead of making a keycloak authenticator, I might make an implicit_oidc authenticator

Issue #1035, opened 2020-12-09
Fact values comparison should be case insensitive


smortex:
Puppet variable comparison is case-insensitive, it's not unusual to see tests like `if $facts.get('os.family') == 'debian'` in a Puppet manifest, while the actual fact value is `Debian`.

When using mcollective (and now choria), the user has to keep in mind the variables case, `Debian` in the previous example:

```sh-session
romain@marvin ~ $ # Good examples:
romain@marvin ~ $ mco find -F os.family=Debian
romain@marvin ~ $ mco find -F os.family=FreeBSD
romain@marvin ~ $ mco find -F os.family=windows
romain@marvin ~ $ # Bad examples:
romain@marvin ~ $ mco find -F os.family=debian
romain@marvin ~ $ mco find -F os.family=freeBSD
romain@marvin ~ $ mco find -F os.family=Windows
```

By using case-insensitive string comparison, we can be more forgiving regarding these kind of inconsistencies with the end user.

ripienaar:
very good, thank you :)

PR #1036, opened 2020-12-09
(#1035) Do case-insensitive fact value comparison

smortex:
Align with the Puppet way of comparing strings, that is ignore the case for comparison.

I see no test for this particular area of the code, but since I am probably not doing the right thing to run them I tested on my laptop by replacing the binary and it seems to be good.  Can't I test with:

```sh-session
romain@zappy ~/Projects/go-choria % go test
?   	github.com/choria-io/go-choria	[no test files]
romain@zappy ~/Projects/go-choria % echo $?
0
```


smortex:
Similarly, it's possible to `include profile::base::${facts.get('os.family')}` in a puppet manifest, and the included class is `profile::base::debian` when `os.family=Debian` (switched to downcase).

Do you want me to adjust the class filter accordingly?

ripienaar:
Yeah that would be good

smortex:
:ballot_box_with_check: Done!

smortex:
Oh, I was not sure about how to handle `reMatch`.  It looks like go does not have a flag to do a case insensitive matching of regexp, and I am not confident enough to add `(?i)` at the beginning of the value…  What do you think about this?

ripienaar:
To test you can either do `go test ./...` or `ginkgo -r .`

filter/classes/classes.go
```diff
@@ -63,7 +63,7 @@ func hasClassMatching(needle string, stack []string) bool {
 
 func hasClass(needle string, stack []string) bool {
 	for _, class := range stack {
-		if class == needle {
+		if class == strings.ToLower(needle) {
```
ripienaar, line 66: maybe do `lneedle := strings.ToLower(needle)` above the for loop then use `if class == lneedle`

ripienaar (changes requested):
looks great, one tiny improvement to be quicker and more efficient

ripienaar:
> Oh, I was not sure about how to handle `reMatch`. It looks like go does not have a flag to do a case insensitive matching of regexp, and I am not confident enough to add `(?i)` at the beginning of the value… What do you think about this?

looks like prepending that `(?i)` is about the only way it can be done

filter/classes/classes.go
smortex: Done!

smortex:
Commits adjusted and rebased, thanks!

ripienaar (approved):

ripienaar (changes requested):
We should add some tests I think, should be easy on the existing tests

ripienaar (approved):

PR #1057, opened 2021-01-02
Fake the use of DNS sans

vjanelle:
If a DNS san doesn't exist, add one which contains the common name.

### DIFF
--- a/config/choria.go
+++ b/config/choria.go
@@ -94,6 +94,7 @@ type ChoriaPluginConfig struct {
 	FileSecurityKey         string `confkey:"plugin.security.file.key" type:"path_string"`         // When using file security provider, the path to the private key
 	FileSecurityCA          string `confkey:"plugin.security.file.ca" type:"path_string"`          // When using file security provider, the path to the Certificate Authority public certificate
 	FileSecurityCache       string `confkey:"plugin.security.file.cache" type:"path_string"`       // When using file security provider, the path to the client cache
+	FileSecurityAppendSAN   bool   `confkey:"plugin.security.file.append_san" default:"true"`      // Assume we are using puppet based certificates, without a SAN, and append one
 
 	CertManagerSecurityNamespace  string   `confkey:"plugin.security.certmanager.namespace" default:"choria"`   // When using Cert Manager security provider, the namespace the issuer is in
 	CertManagerSecurityIssuer     string   `confkey:"plugin.security.certmanager.issuer"`                       // When using Cert Manager security provider, the name of the issuer
--- a/providers/security/filesec/file_security.go
+++ b/providers/security/filesec/file_security.go
@@ -14,6 +14,8 @@ import (
 	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/asn1"
 	"encoding/json"
 	"encoding/pem"
 	"errors"
@@ -88,6 +90,9 @@ type Config struct {
 
 	// TLSSetup is the shared TLS configuration state between security providers
 	TLSConfig *tlssetup.Config
+
+	// Should we auto-append a SAN if it is not present?
+	AppendSAN bool
 }
 
 type signingRequest struct {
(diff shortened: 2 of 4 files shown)


ripienaar:
this is very cool if it works, I somehow assumed we cant modify the cert :)

anyway yeah this looks good, clean it up and we can go with this.  I'd add a big comment block to say wtf we do this crazy thing

vjanelle:
It's non-intuitive, for example you have to also append the CN to DNSNames because x509/verify will populate that later.

.circleci/config.yml
```diff
@@ -32,7 +33,7 @@ jobs:
 
   build_el6_64bit_rpms:
     docker:
-      - image: circleci/golang:1.14
+      - image: circleci/golang:1.15
```
vjanelle, line 36: Just testing for this build, will remove later.

vjanelle:
@ripienaar messed something up with the packager name generation, will need to look at that tomorrow.

.circleci/config.yml
```diff
@@ -30,16 +31,16 @@ jobs:
...
+  build_el7_64bit_rpms:
     docker:
-      - image: circleci/golang:1.14
+      - image: circleci/golang:1.15
 
     working_directory: /go/src/github.com/choria-io/go-choria
 
     environment:
-      PACKAGE: el6_64
+      PACKAGE: el7_64
```
ripienaar, line 41: as on slack I think we need to keep el6 a bit longer

providers/security/filesec/file_security.go
```diff
@@ -534,6 +539,54 @@ func (s *FileSecurity) VerifyCertificate(certpem []byte, name string) error {
...
+		if e.Id.Equal(oidExtensionSubjectAltName) {
+			hasSan = true
+		}
+	}
+
+	if !hasSan && s.conf.AppendSAN {
+		s.log.Debugf("Certificate for %s does not have a SAN, appending one", cert.Subject.CommonName)
+
+		// We do not have a SAN, so continue down this path of creating one
+		var e error
```
ripienaar, line 571: prefer to use `err`

providers/security/filesec/file_security.go
```diff
@@ -534,6 +539,54 @@ func (s *FileSecurity) VerifyCertificate(certpem []byte, name string) error {
...
+
+		extSubjectAltName.Value, e = asn1.Marshal([]string{"DNS:" + cert.Subject.CommonName})
+
+		// This value is usually populated when the certificate is parsed, this is necessary
+		// since the validation code in x509/verify.go will reference this array vs rebuilding
+		// the list of SANs every time
+		cert.DNSNames = append(cert.DNSNames, cert.Subject.CommonName)
+
+		if e != nil {
+			s.log.Error(e.Error())
```
ripienaar, line 582: some more details here if we were going to log, though probably we can just skip logging all together as the callers will handle the error and log if they wish

.circleci/config.yml
```diff
@@ -63,154 +63,125 @@ jobs:
...
-
-    working_directory: /go/src/github.com/choria-io/go-choria
-
-    environment:
-      PACKAGE: el7_64
-      BUILD: foss
-      BUILDER: choria/packager:el7-go1.14
-
-    steps: *std_build_steps
+            - artifacts
```
ripienaar, line 76: got a extra space here

config/choria.go
```diff
@@ -94,6 +94,7 @@ type ChoriaPluginConfig struct {
 	FileSecurityKey         string `confkey:"plugin.security.file.key" type:"path_string"`         // When using file security provider, the path to the private key
 	FileSecurityCA          string `confkey:"plugin.security.file.ca" type:"path_string"`          // When using file security provider, the path to the Certificate Authority public certificate
 	FileSecurityCache       string `confkey:"plugin.security.file.cache" type:"path_string"`       // When using file security provider, the path to the client cache
+	FileSecurityAppendSAN   bool   `confkey:"plugin.security.file.appendSAN" default:"true"`       // Assume we are using puppet based certificates, without a SAN, and append one
```
ripienaar, line 97: make it `append_san` we have no capitals in config strings at the moment

ripienaar (changes requested):
few minor things, looking good

providers/security/filesec/file_security.go
vjanelle: I don't know how it'd actually fail.  There's no validation on the marshalling as far as I can tell, unless some non-..ascii? data got in there.

ripienaar: maybe :) or they might want to add validation in a future release and didnt want to then change the API?

vjanelle: Right now, the only way that'd get in is if you were somehow able to marshal data to begin with, which would break the ASN1 interpreter.  

Or memory corruption.

vjanelle:
@ripienaar done - tried to add the extensions to the CSR request, but they get removed at some point.

Tried to add circleci/golang:1.15 as an additional test container, but it times out.

Issue #1060, opened 2021-01-06
Golang 1.15 depreciates certificates without CNs


vjanelle:
If a SAN doesn't exist, fake one so that x509.Verify doesn't choke.

Test golang 1.15 concurrently.

PR #1061, opened 2021-01-06
(#1060) Move name validation logic out of x509/Verify

vjanelle:
@ripienaar other alternative I think to #1057


ripienaar (approved):

vjanelle:
The CA validation query with the blank name, wouldn't it be better to isolate the CA checks from the name check?

ripienaar:
> The CA validation query with the blank name, wouldn't it be better to isolate the CA checks from the name check?

it would, but I wouldnt reject the PR without it :)

Issue #281, opened 2018-04-30
improve windows support


ripienaar:
- [ ] do not log in color to the terminal
 - [ ] have a real configuration dir where config files go, now uses \etc...
 - [x] add a windows compatible service runner thing ([example](https://github.com/golang/sys/blob/master/windows/svc/example/service.go))

ripienaar:
start with:

```
C:\Program Files\Choria\bin
     choria.exe
     choria.bat
     choria_mcollective_agent_compat.rb # from the module

C:\Program Files\Choria\log
     choria*.log

C:\ProgramData\Choria\
     server.conf
     broker.conf
```

netdata-be:
Hi there, can I assist you with the windows support?
We spoke a CfgMgmtCamp last month, and would really like to get choria running for windows.

ripienaar:
Hey, thanks a lot for the offer!

I am scheduled to start digging into what it means to get this going in a week or 3, let me get back to you?

Meanwhile I can try document how to build a windows binary and you can try it and see where things break and start opening issues around that? Then we just have to itterate till things are functional I think.

Apart from that I need to understand what would make sense for packing, a msi or chocolatey or is a msi a prereq for chocolaty etc?

ripienaar:
```
% docker run -ti -v `pwd`:/tmp/base --rm choria/packager:el7-go1.10 bash
# yum -y install git
# mkdir -p /go/src/github.com/choria-io
# cd /go/src/github.com/choria-io/
# git clone https://github.com/choria-io/go-choria.git
# cd go-choria/
#  go get github.com/Masterminds/glide
# /go/bin/glide install
.....
[INFO]  Replacing existing vendor dependencies
# go generate
2019/02/25 16:32:16 Generating plugin loaders from packager/plugins.yaml
2019/02/25 16:32:16 Generating loading code for plugin golangmco from github.com/choria-io/mcorpc-agent-provider/mcorpc/golang into plugin_1_golangmco.go
2019/02/25 16:32:16 Generating loading code for plugin rubymco from github.com/choria-io/mcorpc-agent-provider/mcorpc/ruby into plugin_2_rubymco.go
# GOOS=windows GOARCH=amd64 go build
# ls -l go-choria.exe
-rwxr-xr-x 1 root root 20528128 Feb 25 16:32 go-choria.exe
# cp go-choria.exe /tmp/base
# exit
```

[...]

PR #1066, opened 2021-01-07
(#281) Allow running the client on windows

smortex:
This PR make choria look after the proper client configuration file on windows.

<details>
<summary>Details from a now fixed issue, see linked PR</summary>
For now, this only change the class error returned to the user from:

```
Could not configure Choria: ping failed to configure: could not parse configuration: open /etc/puppetlabs/mcollective/client.cfg: The system cannot find the path specified.
```

to:

```
Could not configure Choria: ping failed to configure: could not parse configuration: open C:\\ProgramData\\choria\\etc\\client.conf: Access is denied.
```

I am a bit clue-less with windows files permissions…  Opening a PR anyway so that people can find it and maybe provide hints to fix this issue!
</details>


smortex:
Current status:

1. `choria enroll` wants to generate a certificate matching the node FQDN and not `%USERNAME%.mcollecive` as it does on other systems.  I worked around this issue by passing `--certname foo.mcollective` to `choria enroll`;  Because of this, I also had to `set MCOLLECTIVE_CERTNAME=foo.mcollective` for the rest of my tests;
2. `choria ping` worked as expected \o/
3. `choria req service status foo -C profile::foo` resulted in a lot of errors (exceptions I think but I lost this output) mixed with the expected output and unhandled ANSI escape sequences, so the terminal output with mixed stderr and stdout was hardly readable;
4. `choria req service status foo -C profile::foo` a second time resulted in the expected output and behavior :shrug: 

The progress-bar cause the screen to be cleared (that's why I lost the error messages), and the ANSI escape sequence for output coloring are not being handled by windows terminal, so this is a bit annoying but basically usable, which is good!

I think the point 1 above must be fixed (I would like to add it to this PR).  Maybe 3 was a fluke or bug in some SSL code path (I saw a quite a lot of segfaults when doing my various tests when asking for a certificate… probably missing files, but I could not determine the exact cause).  The rest is mostly cosmetic, so i think I will not handle it as part of this PR.

I'll continue to work on this next week…

smortex:
Point 1 fixed in new commit.  I think it starts to be good for reviewing, I will be able to continue working on this next week.

ripienaar:
Thanks a lot for this. Will need issues I suspect some of these will result in a lot of work 

We do have a setting to disable color I probably should honour that and default to off on windows. Though I was under the impression my color lib supported windows. What terminal did you use? Suspect much more success on the new one

you can disable the progress bar on the CLI with —np to help

ripienaar:
merged this one, did the color fixes you should already be in a much better place I hope, lets see what else crawls out, feel free to just tag these things on #281

PR #1067, opened 2021-01-07
(#1076) Remove trailing dot of SRV target

smortex:
When using SRV records, the canonical hostname of the machine providing the service is expected to end with a dot.

But if this dot is kept, TLS connection will fail because the hostname does not match the Sever Name Indication (SNI):

```
org.eclipse.jetty.http.BadMessageException: 400: Host does not match SNI
```

Remove any trailing dot at the end of the target to fix this issue.

### DIFF
--- a/srvcache/cache.go
+++ b/srvcache/cache.go
@@ -2,6 +2,7 @@ package srvcache
 
 import (
 	"net"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ripienaar:
This is really weird I use SRV all the time. But agree we should remove it. 

Need an issue though this is a bit more than a small thing

ripienaar:
The exception here is from jetty a Java webserver, was this a request to PuppetDB for discovery perhaps?

smortex:
Maybe something changed or was fixed in the latest go (I am compiling with the latest go 1.15 — the one that breaks TLS CN matching), the latest puppetserver (I am using Puppet 7) or the latest Ruby (puppet 7 is bundling Ruby 2.7).

I have also been using SRV records for quite some time myself, and I think I would have been bitten by this too.

ripienaar:
Yeah reading about this it seems the http spec isnt exactly clear - so dropping the `.` is safe for sure

smortex:
I checked the error is from PuppetServer, not PuppetDB (I can see the error in puppetserver's log file and to be sure I tested with PuppetDB stopped).

This is confirmed with AIO PuppetServer 6.14.1-1buster on Debian and puppetserver7-7.0.1 on FreeBSD, so maybe not so new…  I guess that in my case I enrolled using the legacy mcollective tools…

In both case, the above patch fix the issue.

Issue #1068, opened 2021-01-08
Improve windows support for color


ripienaar:
Windows terminals are not great at color, so best default the color support to off on windows.

 * Support it in `choria req`
 * Reply formats
 * Generated clients

PR #1069, opened 2021-01-08
(#1068) disable color on windows, improve color setting support

ripienaar:


ripienaar:
@smortex this should improve the color situation - by just completely removing color for windows

Issue #751, opened 2019-12-19
Add support for "Project directories" for clients


smortex:
Clients need a bunch of files for reaching the middleware, including client configuration, client certificate, client private key and ca certificate. These files are stored in well-known locations (e.g. `~/.mcollective/client.cfg`; `~/.puppetlabs/etc/puppet/ssl/certs/$USER.mcollective.pem`; `~/.puppetlabs/etc/puppet/ssl/privte_keys/$USER.mcollective.pem`; `~/.puppetlabs/etc/puppet/ssl/certs/ca.pem`). A user can use the `--config` flag to point the tools to another configuration file, which in turn can setup a custom `ssldir` to locate certificates in another location.  When dealing with multiple choria infrastructures for multiple customers, it makes sense to easily switch from a configuration to another, and manually adding `--config ~/customer-config/client.cfg` to each command is tedious.

Bolt have a concept of "project directories": when a `bolt` command is run, it looks for a local configuration in the current directory.  If none is found, it checks the parent directory, and so on to the root of the filesystem.  If no configuration was found, the user default configuration is used, and if the user does not have a user configuration, system configuration is used as a last resort.
Bundler has a similar behavior when working with a Ruby project: running bundler from a sub-directory of the project will use the project's bundle.

While using wrappers around standalone commands is something that helps (e.g. https://github.com/smortex/mco_env), it does not scale when working beyond the `mco` command, e.g. when writing custom inventory scripts.

ripienaar:
OK, yeah thats not a bad idea, but we'd have to call this kind of config something else, like just `choria.cfg` no `.`, and they would need a way to include others I think.

So:

```
# /home/you/project1/choria.cfg
include "/etc/choria/client.cfg

plugin.choria.middleware_hosts = blah
```

else you have to set a bunch of other things?

Still the plugin.d directory will screw this up - its relative in location to the config file, so if you use `~/.mcollective` today then `/etc/choria/plugin.d` or whatever isnt read, and of course thats all that puppet manages.

smortex:
I started working on this.

Didn't though about this `include` thing but I think it can be helpful to DRY-up things (Don't Repeat Yourself).  I was initially thinking about taking into account multiple configuration file:  allowing the system to provide a system-wide configuration file, which a user can override settings in their personal configuration file, which projects can override settings in projects configuration files, which sub-projects can override in projects configuration files, etc…

[...]

PR #1071, opened 2021-01-09
(msic) Add path to FreeBSD clientconfiguration file

smortex:
This file is already managed by the Puppet modules, but choria does not look for this path.  Add it to the list of checked configuration files.

I guess I will work on #751 next. It will be in the same area and is a must-have for my way of using mcollective/choria :wink:


ripienaar:
thanks!

PR #1079, opened 2021-01-10
(#751) Add support for project configuration files

smortex:


choria/util.go
```diff
@@ -19,6 +19,54 @@ import (
...
+
+	return files[0]
+}
+
+// ApplicableConfigurationFiles return a list of configuration file to take into account to
+// build the current configuraiton
+func ApplicableConfigurationFiles() ([]string) {
+	var res []string
+
+	res = append(res, ProjectConfigurationFiles(os.Getwd())[:]...)
```
ripienaar, line 35: probably dont need the `[:]`

ripienaar:
this is interesting, I guess we'd need to do it like this though:

```nohighlight
if hasProjectConfig()
  load default user config and dot files
  load ~/.choriarc etc without parsing dot files just overriding
  load project file without parsing dot files just overriding
else
  load default user config as today
end
```
I think thats the only way this is possible but will require quite a lot of surgery to config parsing

choria/util.go
smortex: :+1:

smortex:
> this is interesting, I guess we'd need to do it like this though:
> 
> ```
> if hasProjectConfig()
>   load default user config and dot files
>   load ~/.choriarc etc without parsing dot files just overriding
>   load project file without parsing dot files just overriding
> else
>   load default user config as today
> end
> ```
> 
> I think thats the only way this is possible but will require quite a lot of surgery to config parsing

I do not understand what you mean with _parsing dot files_, can you clarify this point please?

ripienaar:
> > this is interesting, I guess we'd need to do it like this though:
> > ```
> > if hasProjectConfig()
> >   load default user config and dot files
> >   load ~/.choriarc etc without parsing dot files just overriding
> >   load project file without parsing dot files just overriding
> > else
> >   load default user config as today
> > end
> > ```
> > 
> > 
> > I think thats the only way this is possible but will require quite a lot of surgery to config parsing
> 
> I do not understand what you mean with _parsing dot files_, can you clarify this point please?

Usually when choria parses `/etc/choria/server.conf` it also parse all the files in `/etc/choria/plugin.d`, it's kind of janky and now does it for all config files in ~/.choriarc for example thats something to fix.

Anyway we need to only parse these things in the first file - puppet maintains them.

smortex:
Ah understood: indeed project config-files should be "self-contained".

config/config.go
```diff
@@ -173,18 +173,22 @@ func NewConfig(path string) (*Config, error) {
...
 	err := parseConfig(path, c, "", c.rawOpts)
 	if err != nil {
 		return nil, err
 	}
 
-	err = parseConfig(c.ConfigFile, c.Choria, "", c.rawOpts)
-	if err != nil {
-		return nil, err
-	}
-
```
smortex, line 185: The exiting code seems to load the configuration twice.  I remove this second call to `parseConfig()` assuming that was safe, but maybe I am missing something…

config/util.go
```diff
@@ -15,6 +15,39 @@ import (
...
+
+	if FileExist(config) {
+		res = append(res, config)
+	}
+
+	return res
+}
+
+// FileExist checks if a file exist on disk
+func FileExist(path string) bool {
```
smortex, line 41: I copied this function from `choria/util.go` because attempting to use it directly produce circular dependencies.  I am unsure however about if I should remove it from that other location and update the rest of the project to call `config.FileExist()`?

smortex (commented):
After many attempts to reach my goal without success, I finally managed to arrive to this small changeset which achieve the expected goal \o/.

After loading the configuration file the usual way, project configuration files are looked for and loaded before continuing.

I have a few points which I am not sure about however, so I added in-line comments.

config/config.go
smortex: Ahhhh… `c` != `c.Choria` (2nd parameter)… This is breaking the test suite

config/util.go
ripienaar: its ok to duplicate, bloody go annoyances. the one in choria is probably used elsewhere so its fine

config/config.go
```diff
@@ -173,18 +173,34 @@ func NewConfig(path string) (*Config, error) {
 	c := newConfig()
 	c.ConfigFile = path
 
+	log.Debugf("Loading choria configuration from '%s'", path)
```
ripienaar, line 176: I'd remove these 2 log lines, useful when developing but this early on logging is not configured yet and we might end up spewing this to peoples console/stderr at times, it's probably fine but I just try to avoid logging here entirely except for warning/errors

config/config.go
```diff
@@ -173,18 +173,34 @@ func NewConfig(path string) (*Config, error) {
...
 		return nil, err
 	}
 
-	err = parseConfig(c.ConfigFile, c.Choria, "", c.rawOpts)
+	err = parseConfig(path, c.Choria, "", c.rawOpts)
 	if err != nil {
 		return nil, err
 	}
 
 	c.parseAllDotCfg()
```
ripienaar, line 188: well aren't we lucky I split this out already :P

config/util.go
```diff
@@ -15,6 +15,37 @@ import (
 	"github.com/choria-io/go-choria/confkey"
 )
 
+// ProjectConfigurationFiles returns any configuration file in the specified directory and their parents directories.
+func ProjectConfigurationFiles(path string, err error) []string {
```
ripienaar, line 19: hmm, this will search all the way up to `/`? should we somehow restrict this to only directories the user own? Or is going all the way up the intended behaviour? I am not really sure whats best - it's probably harmless but I'm weary of grabbing random stuff from all over the place.  Hard to anticipate all the implications and what kind of abuse a user other than you can do if they can influence your config?

config/config.go
smortex: Yeah, I was concerned that the config can change while parsing the configuration files, but was happy to see that passing `-d` had the expected behavior…  But this piece of information is not very useful, so I'll remove it.

config/util.go
smortex: I also had concerns about this and finally decided that going to the filesystem root was maybe preferable: while usually a had practice, users might be working on project outside of their `$HOME` directory (e.g. from the deployment directory of an application, an NFS share or whatever) and in a Continuous Deployment context `$HOME` might not be set and the deployment user might be rooted in a file hierarchy that does not belong to them (I do have this case, but the CD system setup choria "system-wide" in the CD environment before triggering the deployment, so projects configuration files are not really an issue neither).

My assumption here is that the system is trustworthy and files which are not under the control of the user are as "safe" as the system-wide configuration file…

ripienaar: @vjanelle what do you think?

vjanelle: Doesn't match up well with the LSB,  but CI/CD use cases and containers throw all of that out of the window.

Whoever owns higher level directories most certainly has higher privileges - if this is the server component, that could be a bad thing, but for clients it's not an issue.

vjanelle: File ownership could become a thing - I'd like to see this disabled by default ideally.

config/config.go
```diff
@@ -178,13 +178,25 @@ func NewConfig(path string) (*Config, error) {
...
 
+	for _, path := range ProjectConfigurationFiles(os.Getwd()) {
```
vjanelle, line 188: This should probably check InitiatedByServer == false.

smortex: That definitely makes sense!  Fixing this.

ripienaar:
very good, thanks a lot!

config/config.go
```diff
@@ -180,13 +180,27 @@ func NewConfig(path string) (*Config, error) {
...
 
-	err = parseConfig(c.ConfigFile, c.Choria, "", c.rawOpts)
+	err = parseConfig(path, c.Choria, "", c.rawOpts)
 	if err != nil {
 		return nil, err
 	}
 
 	c.parseAllDotCfg()
 
+	if !c.InitiatedByServer {
```
ripienaar, line 190: There is no way InitiatedByServer can be set at this point since the config was just created in the same function, so we need a rethink, we also need to consider this on the broker to avoid it, so I need to think a bit more about how to approach this.

Issue #1092, opened 2021-01-16
support an external method of discovery


ripienaar:
It should call a configured command and use the results for discovery.

input should be:

```json
{
   "filter": {},
   "collective": "mcollective",
   "timeout": 1
}
```

output should be:

```json
{
   "nodes": ["one"],
   "error": "reason for failure or blank/absent"
}
```

ripienaar:
@vjanelle will this do what you want?

ripienaar:
https://github.com/choria-io/go-choria/pull/1093/files#diff-531ebf02d58a9e3cd1dc31e2d39ab32a77ab674042e97b18c8652c0f7ebed305

ripienaar:
I think for now this is fine, but in time we can probably add support to specify multiple externals in the config and you can `--dm` different external ones, so I'm going to merge this and we can itterate

PR #1093, opened 2021-01-16
(#1092) add an external discovery method

ripienaar:


client/discovery/external/testdata/good.rb
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env ruby
+
+require "json"
+require "pp"
+
+if ARGV.length != 1
```
ripienaar, line 6: this is just to test the arg parsing works, and probably not even a good test, wouldnt be needed normally

ripienaar:
Using this with the `go-external` helper making a new discovery method is basically:

```go
func main() {
	jsonDiscovery := discovery.NewDiscovery(func(ctx context.Context, timeout time.Duration, collective string, filter discovery.Filter) ([]string, error) {
		f, _ := ioutil.ReadFile("/etc/choria/nodes.json")
		nodes := []string{}
		err := json.Unmarshal(f, &nodes)
		
		return nodes, err
	})
	
	jsonDiscovery.ProcessRequest()
}
```

PR #1109, opened 2021-01-23
(#1110) Improve PuppetDB discovery

smortex:
PuppetDB discovery is not working as expected because the default
configuration take precedence on discovery based on SRV records.

Rework this part of the code so that PuppetDB discovery follows this
scheme:

1. If a host is set in the configuration file, use this;
2. If SRV lookup is enabled:
    1. Try the _x-puppetdb._tcp SRV record;
    2. Try the _x-puppet._tcp SRV record and change the port;
3. Use the default host, "puppet".

### DIFF
--- a/choria/framework.go
+++ b/choria/framework.go
@@ -13,7 +13,6 @@ import (
 	"net/http"
 	"net/url"
 	"os"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


ripienaar:
Definitely not a misc, need a issue and we need to see how ruby ones work. My gut feel is that if configured it should use what’s DJ figured rather than be overridden by SRV

smortex:
Reworked to be in-line with MC: always perform SRV lookups, on failure try with another SRV lookup and "fixing" the port, and only if all fail use configured host and port.

ripienaar:
great, will take a close look tomorrow, we might have the same issue elsewhere too

smortex:
I am a bit concerned that SRV records are systematically checked, even if `use_srv=false` and that setting `puppetdb_host` is ignored if some SRV records where found, making it impossible to override them.

It's the way MC works, so I guess that if we want to be backward compatible it makes sense to not change this…  But it looks a bit odd :shrug:

ripienaar:
Definitely should when disabled. I am not against improving things if mco is also broken :)

will have a good look today

choria/framework.go
```diff
@@ -269,25 +268,37 @@ func (fw *Framework) FederationMiddlewareServers() (servers srvcache.Servers, er
 
 // PuppetDBServers resolves the PuppetDB server based on configuration of _x-puppet-db._tcp
 func (fw *Framework) PuppetDBServers() (servers srvcache.Servers, err error) {
-	configured := ""
-	if fw.Config.Choria.PuppetDBHost != "" {
-		configured = fw.Config.Choria.PuppetDBHost
-		if fw.Config.Choria.PuppetDBPort != 0 {
-			configured += ":" + strconv.Itoa(fw.Config.Choria.PuppetDBPort)
-		}
+	defaultHost := fw.Config.Choria.PuppetDBHost
```
ripienaar, line 271: So maybe we do a few things here:

 * We remove the defaults from the config, so puppetdb settings are empty 
 * We change this method to:

``
if puppetdb_set
   return configured

if srv_enabled
   attempt to resolve via SRV
   return if found

return "puppet:1234"
```

I am happy with hard coding these defaults here, this makes it so srv is only done when not set and trusts configured values.

what do you think?

Comment (smortex) [reply choria/framework.go]: :+1: I updated the PR to match this behavior.

File: choria/framework.go [hunk 2]
```diff
@@ -269,25 +268,41 @@ func (fw *Framework) FederationMiddlewareServers() (servers srvcache.Servers, er
...
-	}
+		configured := fmt.Sprintf("%s:%d", fw.Config.Choria.PuppetDBHost, fw.Config.Choria.PuppetDBPort)
 
-	if configured != "" {
 		servers, err = srvcache.StringHostsToServers([]string{configured}, "https")
 		if err != nil {
 			return servers, fmt.Errorf("could not parse configured PuppetDB host: %s", err)
 		}
-	}
+	} else {
```
ripienaar, line 278: you can return servers above the else, then completely remove the else.  This removes the if within an else with more if's in it.  ends up reading a bit easier

ripienaar (changes requested):
looks pretty good, thanks for this.  Just a small ask to make it a bit less nested

choria/framework.go
smortex: Hum, I though some code bellow all the condition was needed in all cases, but it is only useful in the context of SRV records querying.  I moved it next to the SRV lookup, and adjusted the conditions as requested.

PR #1111, opened 2021-01-23
Add bash completion script

optiz0r:
Adds a bash completion script which handles most commands and options using the built-in kingpin `choria --completion-bash` command to generate the possible completions, but for `rpc/req` uses `choria completion` subcommands to populate agents, actions and inputs for rpc.

### DIFF
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -13,12 +13,13 @@ import (
 type completionCommand struct {
 	command
 
-	showZsh   bool
-	showBash  bool
-	list      string
-	agent     string
-	action    string
-	zshScript string
+	showZsh    bool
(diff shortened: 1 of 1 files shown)


cmd/completion.go
```diff
@@ -13,12 +13,13 @@ import (
...
-	agent     string
-	action    string
-	zshScript string
+	showZsh    bool
+	showBash   bool
+	list       string
+	agent      string
+	action     string
+	zshScript  string
+    bashScript string
```
ripienaar, line 22: can you give this a run through `go format`? it'll fix formatting etc

ripienaar:
Thnk you :)

PR #1112, opened 2021-01-24
(misc) Fix display of discovered nodes count in reply footer

smortex:
A regression was introduced in cebe2cd78c63f1260d5198945b9c2cc67626a691
where the number of discovered nodes is not displayed if this number is
0 or if the number of responses is different from the number of
discovered nodes.

Properly initialize the string representation of the number of
discovered nodes to fix this issue.


ripienaar:
thanks

PR #1134, opened 2021-02-02
(misc) Sort classes alphabeticaly in inventory

smortex:
mco(1) used to do this sorting, and it makes the output more readable for humans, so sort the list of configuration management classes.

**Note:** The list of agents above the configuration management classes appears to be sorted but use a similar code.  maybe it's wise to sort it explicitly too?

### DIFF
--- a/filter/classes/classes.go
+++ b/filter/classes/classes.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"sort"
 	"strings"
 )
 
@@ -90,6 +91,8 @@ func ReadClasses(file string) ([]string, error) {
 	for scanner.Scan() {
 		classes = append(classes, scanner.Text())
 	}
+
+	sort.Strings(classes)
 
 	return classes, nil
 }


smortex:
@ripienaar I considered this to be a cosmetic change, but do we need an issue for it?

cmd/inventory.go
```diff
@@ -124,6 +124,7 @@ func (i *inventoryCommand) inventoryAgent() error {
 	fmt.Printf("  Configuration Management Classes:\n\n")
 	longest := util.LongestString(inventory.Classes, 40)
 	format := fmt.Sprintf("    %%-%ds %%-%ds\n", longest, longest)
+	sort.Strings(inventory.Classes)
```
ripienaar, line 127: Lets sort here instead and we solve it for all use cases

https://github.com/choria-io/go-choria/blob/97b23c3eab4ce4810fe35f66d284f2d3141f1068/filter/classes/classes.go#L94

no issue is fine

smortex: Makes sense, but now the client blindly displays the list in the order provided by the server.  That might be surprising in an environment with different versions of choria, but I am perfectly fine with this.

I adjusted the PR accordingly.

ripienaar:
Yeah next release I'll have to bump choria in the puppet modules - rather than the current "present" setting, since the ruby libraries will require this version of choria at least.

So we'll have to bump it all up at the same time, so I think it'll be good

PR #1135, opened 2021-02-02
(#1136) Fit progressbar in a 80 columns terminal

smortex:
When using a terminal with the default 80x25 size, the rendered progress bar line overflows 80 chars and leads to artifacts, especially if the window is resized while the progress bar is "running".

Support for terminal width is on the [TODO list of the uiprogress module](https://github.com/gosuri/uiprogress#todos) so as a workaround, explicitly set the progress bar width to 50 chars.


ripienaar:
Really need issues :) misc are for like minor doc update etc. Not big UX issues

smortex:
lol, I though this was just a cosmetic change :-)  updating…

ripienaar:
Thanks :)

Issue #1154, opened 2021-03-03
Support ppc64le on RedHat 7 and RedHat 8


treydock:
It would be idea to support ppc64le on both RedHat 7 and RedHat 8.

PR #1155, opened 2021-03-03
(#1154) Support building ppc64le for EL7 and EL8

treydock:
Solves #1154 

binaries and RPMs built and verified on my local IBM ppc64le hardware just that the binaries included in the RPM work on that architecture.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,6 +105,19 @@ jobs:
 
     steps: *std_build_steps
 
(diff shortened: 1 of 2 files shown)


ripienaar:
looks good, lets see how nightly goes

You'd need to test quite carefully these binaries on your machines, on el6 now choria built with go 1.16 just panics at start due to go not supporting unsupported kernels anymore, just a headsup

Issue #1250, opened 2021-05-11
Improve readability of data printed in columns


smortex:
When choria output information in columns, items are ordered from left to right, and then from top to bottom.  Comparing with the output of UNIX command like `ls` which order items from top to bottom first and then from left to right.

The latter makes it easier to seek for a value in a long list, so I propose to change the output to match the ordering of commands like `ls`:

Current output of `choria inventory`:
```
  Agents:

    bolt_tasks      choria_util     discovery      
    filemgr         nettest         package        
    puppet          rpcutil         scout          
    service         unix                           

  Tagged Classes:

    apt                                 apt::params                        
    apt::update                         choria                             
    choria::broker                      choria::broker::config             
    choria::broker::service             choria::config                     
    choria::install                     choria::repo                       
    choria::scout_checks                choria::scout_metrics              
    choria::service                     firefox                            
    locales                             locales::params                    
    mcollective                         mcollective::config                
    mcollective::facts                  mcollective::plugin_dirs           
    mcollective_agent_bolt_tasks        mcollective_agent_filemgr          
    mcollective_agent_nettest           mcollective_agent_package          
    mcollective_agent_puppet            mcollective_agent_service          
    mcollective_choria                  mcollective_util_actionpolicy      
    ntp                                 ntp::config                        
    ntp::install                        ntp::service                       
    optiplex-3050.vittoria.pro          profile::base                      
    profile::base::linux                profile::bolt                      
    profile::choria::leaf               profile::choria::server            
    profile::firefox                    profile::mcollective::unix         
    profile::ntp                        profile::python                    
    profile::sshd::client               profile::vim                       
    profile::vittoria::desktop          profile::vittoria::desktop::linux  
    python                              python::config                     
    python::install                     python::params                     
    role::base                          role::desktop                      
    role::vittoria::desktop             role::vittoria::desktop::developper
    settings                            ssh                                
    ssh::client                         ssh::client::config                
    ssh::install                        systemd                            
    systemd::journald                   wormhole                           
    wormhole::package
``` 

Suggested output:

```
  Agents:

    bolt_tasks      nettest         scout          
    choria_util     package         service        
    discovery       puppet          unix           
    filemgr         rpcutil                        

  Tagged Classes:

    apt                                 optiplex-3050.vittoria.pro         
    apt::params                         profile::base                      
    apt::update                         profile::base::linux               
    choria                              profile::bolt                      
    choria::broker                      profile::choria::leaf              
[...]

PR #1251, opened 2021-05-11
(#1250) Print inventory columns vertically

smortex:
Improve output of choria inventory by sorting items "vertically" in a
similar fashion to `ls(1)` or `column(1)`.

Fixes: #1250


ripienaar:
Hmm, some changes to `staticcheck` causing the tests to fail, I think those should be fixed in the main branch would you mind rebasing?

Also if you dont mind change your commit to `(#xxx) blah blah` format.

smortex:
Oh, done/fixed, thanks!

Issue #1277, opened 2021-06-29
Support loading pkcs8 private keys


vjanelle:
Currently if you feed choria a pkcs8 private key, the filesec provider will choke.

We can handle this by detecting if the pkcs1 parser fails, trying again with the pkcs8 parsers, and then continuing on from there.

As an added bonus this helps set us up for non-RSA keys.

PR #1278, opened 2021-06-29
(#1277) Support PKCS8 containers

vjanelle:


vjanelle:
Also need to figure out how to make a pkcs8 key too.  Woo.

smortex (approved):

vjanelle:
Apparently "openssl pkcs8" was the answer to that one.

vjanelle:
@smortex @ripienaar I think this is as good as it's going to get for RSA support.  I've got another branch with ecdsa/ed25519 support, but that's trickier - the methods aren't the same, in particular for signature validation.

ripienaar:
nice, lgtm. Where do we come across pkc8 keys in the wild?

vjanelle:
@ripienaar Probably only within my testing harness at work.

vjanelle:
@ripienaar would also be important as a step to allowing things other than RSA keys.

PR #1371, opened 2021-08-23
Configure Debian Bullseye build

fklajn:
Signed-off-by: Franciszek Klajn <fklajn@opera.com>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -308,7 +308,7 @@ jobs:
             done
           done
 
-          for d in stretch_64 buster_64 buster_armhf
+          for d in stretch_64 buster_64 buster_armhf bullseye_64
           do
             for i in $(find /tmp/workspace/artifacts/foss/${d} -name \*.deb)
             do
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -211,6 +211,19 @@ jobs:
 
     steps: *std_build_steps
 
+  build_bullseye_64bit_debs:
```
ripienaar, line 214: also need to include it down by `test_and_ship`

fklajn: Included, thanks.

ripienaar:
Also by `gather_artifacts`

ripienaar:
Also please rebase against main

fklajn:
Added Bullseye to gather_artifacts (and to for loop in release_packagecloud) and rebased against master.

ripienaar:
Unfortunately this is broken, you specifically removed the dh-systemd dependency as its merged in the main dhelper which does not exist on bullseye however the packaging does require dh-systemd.  So this is entirely broken.

We will need to create a bullseye specific package template that adjusts this.

https://github.com/choria-io/go-choria/blob/d2bc0b87874afd175e3a674cd14a99c02e9fb98e/packager/templates/debian/generic/control#L7

I am trying to get the release out but it wont have bullseye packages

fklajn:
What about this? https://github.com/choria-io/go-choria/pull/1376

PR #1376, opened 2021-08-25
Fix build dependencies for Debian Bullseye

fklajn:
Signed-off-by: Franciszek Klajn <fklajn@opera.com>

PR #1391, opened 2021-09-01
(maint) Rework Debian 11 packaging

smortex:
The generic and bullseye packaging only differ in a build-dependency,
dh-systemd, which is not shipped anymore with Debian 11.

This package however appears to be a transition package which install
debhelper, the other build dependency we already have (since at least
Debian 8 [1]).

Removing the transition package from the list of build dependencies
should therefore be enough to build a Debian 11 package of choria.

Reverting #1376

[1]: https://packages.debian.org/stretch/dh-systemd


smortex:
@fklajn may I have a review for this?

fklajn:
This is true since Debian 9 (Stretch). As Debian 8 (Jessie) is not supported here (there is no configuration for package build for this release) thus I would vote for removing dh-systemd dependency. Thanks for spotting this!

ripienaar:
Are we sure non of the Ubuntu we support needs this dependency? Else yes good to have one again

ripienaar:
We have Xenial still but its now on security-fixes-only stage so I would love to axe it, but alas we have had over 1k downloads of its packages across the last 2 releases (excluding those who self host packages)

fklajn:
The package is not needed on Bionic, but still needed on Xenial. Maybe the separate template should be created for Xenial instead of Bullseye then?

smortex:
Thanks for the feedback!  I adjusted the PR accordingly.

I refrained from integrating the changes in #1392 but I think they would make sense too.

ripienaar:
great thanks, lets add one of each kind of debian to the nightlies and see it go through all the way

PR #1392, opened 2021-09-02
(#1388) Restart choria-server on Debian on update

### LINKED ISSUE #1388: choria-server does not restart during package upgrade on Debian
When upgrading choria on Debian, the `choria-server` service is not restarted (I could not check this for `choria-broker`).

So while the catalog applied successfully and no more change happen on subsequent runs of Puppet, the old service is still running and has to be restarted manually.

While we can probably do this in the puppet code to notify the service when the package is installed / upgraded, it feels like restarting the service also make sense for people using choria outside of a Puppet environment and should probably be fixed at the package level.

smortex:
So, it happen the logic that can "almost" manage the services as intended is already there:

https://github.com/choria-io/go-choria/blob/00825d7b9dec655577d8bca2f46f8c891417b78a/packager/templates/debian/generic/rules#L30-L34

…just "disabled" by some overrides a few lines above:

https://github.com/choria-io/go-choria/blob/00825d7b9dec655577d8bca2f46f8c891417b78a/packager/templates/debian/generic/rules#L14-L16

I write "almost" because when passing `--no-start` to `dh_installinit` like is done for the broker service, the service is not restarted on upgrade not stopped when the package is removed.

Puppet seems to be quite creative for handling the puppet service manually in their [postinst](https://github.com/puppetlabs/vanagon/blob/main/resources/deb/postinst.erb#L9), [prerm](https://github.com/puppetlabs/vanagon/blob/main/resources/deb/prerm.erb#L21) and [postrm](https://github.com/puppetlabs/vanagon/blob/main/resources/deb/postrm.erb#L19) scripts in order to not start the puppet agent when installing, but restart it when upgrading, stopping it when removing the package and so on.

~~I am not sure if it is wiser to replicate such a setup or just leave the broker service untouched on upgrades.~~  **EDIT** It's not at all what we spoke about in #1388 so I will definitively add this logic :stuck_out_tongue_closed_eyes:  **EDIT2** Done :ballot_box_with_check: 

Fixes #1388

### DIFF
--- a/packager/templates/debian/generic/postinst
+++ b/packager/templates/debian/generic/postinst
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+action="$1"
+
+if [ "$action" = configure ]; then
+	old_version="$2"
+
+	if [ -z "$old_version" ]; then
+		systemctl enable choria-server.service
+	else
+		systemctl try-restart choria-broker.service choria-server.service
+	fi
+fi
--- a/packager/templates/debian/generic/postrm
+++ b/packager/templates/debian/generic/postrm
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+systemctl daemon-reload
--- a/packager/templates/debian/generic/prerm
+++ b/packager/templates/debian/generic/prerm
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+action="$1"
+
+if [ "$action" = remove ]; then
+	systemctl --no-reload disable choria-server.service choria-broker.service
+	systemctl stop choria-server.service choria-broker.service
+fi
--- a/packager/templates/debian/generic/rules
+++ b/packager/templates/debian/generic/rules
@@ -11,10 +11,6 @@ override_dh_auto_test:
 
 override_dh_auto_build:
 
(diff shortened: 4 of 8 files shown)


smortex:
Seems to work as described in #1388

Note however that when building a debian package on my system, the deb file has an empty `/usr/bin/choria`.  Amusingly, the built binary was aside the .deb file so I do not known what is going wrong, but since the mirror packages are fine I think the problem is on my side.

Nevertheless, I checked that the expected commands where invoked:
* When installing, only `choria-server` is enabled, but none of the services are started;
* On upgrade, all running services are restarted (`try-restart`);
* On removal, all services are stopped.

The build failure is unrelated, already addressed in b05b4c61711008b8a6df231f9eaaf61eaff4f267 which is part of another PR.

ripienaar:
Seems good to me, though knowing nothing about debian packages :) merge when ready

neomilium (approved):

smortex:
Tweaked a few things following recommendations from @neomilium.  I tested this on Ubuntu Xenial since this docker environment seems to be working correctly on Debian Bullseye :trollface:

Issue #1395, opened 2021-09-03
Systemd output type syslog is deprecated and should be removed


bastelfreak:
the syslog output within system units is deprecated and newer versions complain about it. The entries `StandardOutput=syslog` and `StandardError=syslog` should be removed. Systemd yells:
```
Standard output type syslog is obsolete, automatically updating to journal. Please update your unit file, and consider removing the setting altogether
```

PR #1396, opened 2021-09-03
(#1395) systemd: Drop legacy syslog logging

bastelfreak:
fixes https://github.com/choria-io/go-choria/issues/1395

### DIFF
--- a/packager/templates/debian/legacy/server.service
+++ b/packager/templates/debian/legacy/server.service
@@ -3,8 +3,6 @@ Description=The Choria Orchestrator Server
 After=network.target
 
 [Service]
-StandardOutput=syslog
-StandardError=syslog
 User=root
 Group=root
 ExecStart={{cpkg_bindir}}/{{cpkg_name}} server --config={{cpkg_etcdir}}/server.conf
--- a/packager/templates/debian/generic/broker.service
+++ b/packager/templates/debian/generic/broker.service
@@ -3,8 +3,6 @@ Description=The Choria Orchestrator Network Broker
 After=network.target
 
 [Service]
-StandardOutput=syslog
-StandardError=syslog
 LimitNOFILE=51200
 User=root
 Group=root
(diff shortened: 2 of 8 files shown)


packager/templates/debian/legacy/server.service
```diff
@@ -3,8 +3,6 @@ Description=The Choria Orchestrator Server
 After=network.target
 
 [Service]
-StandardOutput=syslog
-StandardError=syslog
```
ripienaar, line 7: the `legacy` package is for circa 2016 ubuntu, so we need to be sure this is fine, @smortex might know

smortex: I did this in 0f0f91bb3ce4057e4773e4b6a8fda6a132ea68a9 part of #1388 and which I need to rebase on master :-)

It was already obsolete back at this time, just not warning about it.

smortex (approved):

packager/templates/debian/legacy/server.service
smortex: Ah crap, this PR is also not up-to-date with the base branch, I was ready to merge this one before rebasing my changes.

smortex: @bastelfreak I removed it from my branch since you address all platforms and I only checked the Debian based ones :-)

bastelfreak: I rebased it

smortex:
Meh CI failed.  CI on master also failed. We seem to have a bunch of race conditions in the test suite: here the failure is related to a splay parameter, on master it's related to something that passed while the CI expect it to fail because of a broken pipe…

I am trying to reproduce the errors here to maybe provide a fix, but no luck so far.

Maybe you can amend and force push until CI pass and I can merge :clown_face:

ripienaar:
That’s annoying I have never seen that one fail. We can merge as is

ripienaar:
The anoyance is that CircleCI has started putting us on ever smaller machines and it just doesnt have the resources so various things timeout and go weird :(

ripienaar:
@smortex green now, you can merge :)

PR #1556, opened 2022-01-07
(misc) Create chocolatey package

mtancoigne:
Hello,

We started to work on a Chocolatey package to install Choria on Windows. These are the file to generate a valid `.nupkg` file, but we don't know how to integrate this in the current build process.

Can we speak about it and see if and how it is possible to push updates to the Chocolatey public repository ?

### DIFF
--- a/chocolatey/ReadMe.md
+++ b/chocolatey/ReadMe.md
@@ -1,0 +1,134 @@
+﻿## Summary
+How do I create packages? See https://docs.chocolatey.org/en-us/create/create-packages
+
+If you are submitting packages to the community feed (https://community.chocolatey.org)
+always try to ensure you have read, understood and adhere to the create
+packages wiki link above.
+
+## Automatic Packaging Updates?
+Consider making this package an automatic package, for the best
+maintainability over time. Read up at https://docs.chocolatey.org/en-us/create/automatic-packages
+
+## Shim Generation
+Any executables you include in the package or download (but don't call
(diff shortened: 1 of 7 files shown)


mtancoigne:
Note: the MSI installer was not added to the files and should be placed in `tools`.

neomilium:
@ripienaar Where do you build MSI installer?
In code, it seems talking about the `.exe` but not `.msi`.

ripienaar:
At present the output from the `packager` directory is a .bat file that needs running on windows to make the .msi.  This is not automated at present. 

So getting this to actually build correctly and automatically needs integrating the choco stuff into the `packager` system and a set of GH Actions written to do the windows stuff.  When I did the build automation windows builders werent an option so possible now, but likely a fair bit of work.

I am back working next week.

ripienaar:
If you run, on a machine with docker, the following:

```
PACKAGES=windows_64 rake build 
```

It will build the following:

```
choria-0.0.0-windows-amd64.exe
Choria-0.0.0-windows-x86_64-windows.tgz
```

The 2nd file has all the things you need to run on a windows machine with the wix toolset installed to build a msi.
 
This is based on `packager/buildspec.yaml` and `packager/templates/windows`. We can adjust things so this runs on a windows builder in GHA and probably automate the building of the msi by extending things. But it's all very slanted to running under docker, I am not sure if there are windows containers?

Quite a bit of work here to automate this end to end I am afraid.

Issue #1405, opened 2021-09-08
Provide 64-bit ARM packages (arm64)


grandich:

ripienaar:
For what distros?

grandich:
We are interested in Ubuntu 20 (focal)

ripienaar:
You can add these via a PR:

First we describe how to compile a binary

https://github.com/choria-io/go-choria/blob/8362ec405315d5f0e0a3f8776b93313a6b051ca9/packager/buildspec.yaml#L42-L46

Then we describe how to make debs:

https://github.com/choria-io/go-choria/blob/8362ec405315d5f0e0a3f8776b93313a6b051ca9/packager/buildspec.yaml#L176-L179

And you can then test locally using something like:

```
PACKAGES=xenial_64 rake build
```

Once you get that going send a PR and I'll do the integration with CircleCI and get nightlies built so we can test it

ripienaar:
Please try https://apt.eu.choria.io/misc/choria_0.99.0.20220119_arm64.deb and see if that works for you? I dont have a proper deb repo for nightlies but this one been through the whole tool chain now so a manual test would be great.

Thanks for your help @ananace !

ananace:
Just as a point of verification;
```console
root@pi-stack-3:/tmp# uname -m
aarch64
root@pi-stack-3:/tmp# cat /sys/firmware/devicetree/base/model; echo
Raspberry Pi 4 Model B Rev 1.1
root@pi-stack-3:/tmp# cat /etc/debian_version 
11.2
root@pi-stack-3:/tmp# curl -LO https://apt.eu.choria.io/misc/choria_0.99.0.20220119_arm64.deb 
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 17.8M  100 17.8M    0     0  14.3M      0  0:00:01  0:00:01 --:--:-- 14.3M
root@pi-stack-3:/tmp# dpkg -i choria_0.99.0.20220119_arm64.deb 
(Reading database ... 59160 files and directories currently installed.)
Preparing to unpack choria_0.99.0.20220119_arm64.deb ...
Unpacking choria (0.99.0.20220119) ...
Setting up choria (0.99.0.20220119) ...
root@pi-stack-3:/tmp# choria --version
0.99.0.20220119
```

[...]

PR #1579, opened 2022-01-18
(#1405) Add 64-bit ARM package

ananace:
As I only have access to bullseye systems for testing, I've only written up the buildspec for one of those.
The package builds, installs, and runs fine on my test system though. (a Raspberry Pi 4 Model B)

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -25,6 +25,16 @@ jobs:
           packager_tag: el8-go1.17
           version: nightly
 
+  bullseye_aarch64:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Build
+        uses: choria-io/actions/packager@main
+        with:
+          build_package: bullseye_aarch64
+          packager_tag: bullseye-go1.17
+          version: nightly
+
   linux_tarball:
     runs-on: ubuntu-latest
     steps:
@@ -69,6 +79,7 @@ jobs:
     needs:
       - el7_64
       - el8_64
+      - bullseye_aarch64
       - linux_tarball
       - windows_zip
       - darwin_amd64_tarball
--- a/packager/buildspec.yaml
+++ b/packager/buildspec.yaml
@@ -54,6 +54,13 @@ foss:
         - rm plugin_*.go || true
         - GOOS=linux GOARCH=amd64 go generate --run plugin
 
+    aarch64_linux:
+      os: linux
+      arch: arm64
+      pre:
+        - rm plugin_*.go || true
+        - GOOS=linux GOARCH=amd64 go generate --run plugin
+
     ppc64le_linux:
       os: linux
       arch: ppc64le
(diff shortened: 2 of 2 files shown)


ripienaar:
Nice, if you wish you can pop it into here https://github.com/choria-io/go-choria/blob/main/.github/workflows/nightly.yaml and you'll see nightly builds tomorrow in the nightly repos so we can get a nice end to end test going?

packager/buildspec.yaml
```diff
@@ -54,6 +54,13 @@ foss:
         - rm plugin_*.go || true
         - GOOS=linux GOARCH=amd64 go generate --run plugin
 
+    aarch64_linux:
+      os: linux
+      arch: arm64
+      pre:
+        - rm plugin_*.go || true
+        - GOOS=linux GOARCH=amd64 go generate --run plugin
```
smortex, line 62:
```suggestion
        - GOOS=linux GOARCH=aarch64 go generate --run plugin
```
Typo?

smortex, line 62: Or maybe (I should install go on this machine te tyr):

```suggestion
        - GOOS=linux GOARCH=arm64 go generate --run plugin
```

smortex:
```suggestion
        - GOOS=linux GOARCH=arm64 go generate --run plugin
```

ripienaar: The build will run in a amd64 container but cross compile for aarch64, so we need to do `go generate` with the arch of the container - see arm examples also

smortex: Good catch… So do we need to set GOOS/GOARCH then?  If the build script is run on ARM it will cause trouble?

ripienaar: When the packager runs the builder container those variables are already set to allow cross compilation to arm

but go generate has to run on tbe actual builder container using its actual architecture. So setting these two here temporarily override the ones set by the packager to facilitate cross compiling.

.github/workflows/nightly.yaml
```diff
@@ -25,6 +25,16 @@ jobs:
...
           version: nightly
 
+  bullseye_aarch64:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Build
+        uses: choria-io/actions/packager@main
+        with:
+          build_package: bullseye_aarch64
+          packager_tag: el8-go1.17
```
ripienaar, line 35: bullseye-go1.17 Here

ripienaar: Also add it to the dependency list a bit lower in tbe file

ananace: Every other job seems to use the `el*-go1.17` tag though? Should the bullseye build be unique in that regard?

ripienaar: every other job builds rpms - this one will build a deb to need the deb stuff.

ananace: Even the Windows zip and the MacOS jobs use the `el8-go1.17` tag though, that's my confusion

ripienaar: the go build can happen on anything - go cross compiles from anywhere to anywhere. Those other jobs just need zip/tar nothing OS specific.

But to build a deb we need debian machine so that swaps it over to debian and then all the things to make debs are there.

I will kick a nightly now and we can see what happens, tbf, this stuff is confusing, i hate CI :)

PR #1584, opened 2022-01-19
(#1405) Expand 64-bit ARM buildspec list

ananace:
Now includes aarch64 for the non-EoL versions of Debian and Ubuntu
(Except for Debian Stretch, as its EoL date is in June this year)


ripienaar:
thank you

Issue #1604, opened 2022-02-24
Choria-0.25.0 does not start on Windows


smortex:
The latest choria release fails to start on windows.

PR #1605, opened 2022-02-24
(#1604) Fix startup on windows

smortex:
Some refactoring in e2a6908c6d0c204cf71f6ebe090d04998c7b6046 unintenitonaly replaced a call to `svc.IsAnInteractiveSession()` with `svc.IsWindowsService()` without changing the variable name nor the logic. As a result when choria is started as a service it acted as a regular command, failing to register to the windows service manager and telling it it started successfuly, and when started from the CLI it failed to start because it could not do such registration only accessible to services.

Rename the variable and fix the logic to unbreak windows.

### DIFF
--- a/cmd/server_run_windows.go
+++ b/cmd/server_run_windows.go
@@ -57,12 +57,12 @@ loop:
 }
 
 func (r *serverRunCommand) platformRun(wg *sync.WaitGroup) (err error) {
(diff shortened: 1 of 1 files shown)


ripienaar (approved):
I remember thinking it was weird but godoc suggested it would be fine … guess not. 

Thanks for debugging this

PR #1653, opened 2022-04-18
experimental matchers for integration testing

ripienaar:

### DIFF
--- a/integration/suites/rpcutil/rpcutil_test.go
+++ b/integration/suites/rpcutil/rpcutil_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/choria-io/go-choria/choria"
 	"github.com/choria-io/go-choria/client/rpcutilclient"
 	"github.com/choria-io/go-choria/config"
+	. "github.com/choria-io/go-choria/integration/matchers"
 	"github.com/choria-io/go-choria/integration/testbroker"
 	"github.com/choria-io/go-choria/integration/testutil"
 	"github.com/choria-io/go-choria/providers/agent/mcorpc"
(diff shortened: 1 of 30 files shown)


integration/suites/rpcutil/rpcutil_test.go
```diff
@@ -137,7 +137,7 @@ var _ = Describe("rpcutil agent", func() {
 		It("Should fetch correct collective info", func() {
 			res, err := rpcutilClient.OptionTargets([]string{"localhost"}).CollectiveInfo().Do(ctx)
 			Expect(err).ToNot(HaveOccurred())
-			Expect(res.Stats().OKCount()).To(Equal(1))
+			Expect(res).To(HaveOnlySuccessfulResponses())
```
ripienaar, line 140: @vjanelle this is what I was talking about on slack, give me some more ideas of what you want to be testable and I can see.

Ultimately I want to expand this out enough that I can stand up a infra using docker compose and write integration tests against that asserting things like did all servers get correctly provisioned etc

ripienaar: Other things I am just kind of thinking about - for the context of testing a deployed infra

 * HaveDiscoverableNodes(fw, filter, count) - discovers nodes matching filter
 * OutputWithValue(output, key, value) - takes any generated client result looking for key and val, those found with `res.AllOutputs()`, I might add some accessors to find more like by name or something
 * AllRepliesWithValue(res, key, value) - loops all the outputs and checks they all have the same key and value

ripienaar: An alternative approach would be to add an `assert` package to each generated client with things like:

```go
package assert

import (
	"testing"

	"github.com/choria-io/go-choria/client/rpcutilclient"
)

type StatsProvider interface {
	Stats() rpcutilclient.Stats
}

func HaveUnexpectedResponders(t *testing.T, res StatsProvider, count int) {
	ur := len(res.Stats().UnexpectedResponseFrom())
	switch {
	case count < 0:
		if ur == 0 {
			t.Fatalf("Expected >0 unexpected responders")
		}
	case count == 0:
		if ur > 0 {
			t.Fatalf("Expected no unexpected responders")
		}
	default:
		if ur != count {
			t.Fatalf("Expected %d unexpected responders but had %d", count, ur)
		}
	}
}
```

```go
res, _ := rpcutilClient.AgentInventory().Do(ctx)
assert.HaveUnexpectedResponders(t, res, 0)
```

This would be usable in any testing framework for go and less casting and weird stuff since it would just use the package internals? 

I could even put this in the package directory as `AssertHaveUnexpectedResponders()` (ie. not on the instance) since then it would have access to the private variables in the client, I think I quite like this it avoids having to pollute the clients with weird accessors

PR #1657, opened 2022-04-22
(misc) Create a agent testing harness

ripienaar:
This adds a harness that can be used to host
agents in a running Server instance that has
their actions stubable using gomock


integration/suites/harness/agent_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+		_, rpcutilClient, err = createRpcUtilClient()
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	AfterEach(func() {
+		mockCtl.Finish()
+	})
+
+	Describe("action stubbing", func() {
```
ripienaar, line 131: @vjanelle the beef of things here, after in the BeforeEach() adding the test harness into the server pretending to be rpcutil we call it here using the unmodified generated rpcutil client and doing mocks on that `ping` action.

integration/suites/harness/agent_test.go
```diff
@@ -0,0 +1,152 @@
...
+		It("Should do the correct stubs", func() {
+			rpcutilAgent.Stub("ping", func(_ context.Context, _ *mcorpc.Request, reply *mcorpc.Reply, _ *mcorpc.Agent, _ inter.ConnectorInfo) {
+				reply.Data = map[string]interface{}{
+					"pong": time.Now().Unix(),
+				}
+			}).AnyTimes()
+
+			for i := 0; i < 5; i++ {
+				res, err := rpcutilClient.OptionTargets([]string{"localhost"}).Ping().Do(ctx)
```
vjanelle, line 140: Only thing I couldn't figure out, I tried OptionsFactFilter (after populating testdata/facts.yaml)  and it wouldn't discover any results.

ripienaar:
```
		da, err := discovery.New(srv.AgentManager())
		Expect(err).ToNot(HaveOccurred())
		err=srv.AgentManager().RegisterAgent(ctx, "discover", da, srv.Connector())
		Expect(err).ToNot(HaveOccurred())
```

should sort it, add to `startServerInstance()`

ripienaar: The way discovery works is a filtered message is sent to the `discovery` agent with just "ping", and if the server filter validation worked a "pong" is sent back.

https://github.com/choria-io/go-choria/blob/dcc5c3fdb8bbaf2eb736c50c28325e47755afd59/providers/agent/mcorpc/golang/discovery/discovery.go#L57-L72

Agents dont need to be "RPC" agents as such, any payload is supported in any format and any serialization. RPC agents formalize the payload to make a RPC request, but discovery doesnt need all that so just string ping.

vjanelle (approved):

PR #1658, opened 2022-04-25
(Misc) return meanigful error messages from filesec

vjanelle:


smortex (approved):
Looks good!

vjanelle:
There's one use case that's possible to trigger where error is nil- need to track that down.

ripienaar (approved):
LGTM

Issue #1665, opened 2022-05-10
create a application like system


ripienaar:
mco applications were ruby plugins that was used to build CLI tools, primarily they were used to render results differently and less frequently about doing multiple chained requests (puppet runall)

For a large portion of things really all thats needed now is a way to construct complex `choria req` commands. 

I imagine that:

```
someapp machine report check_load --lt 1.0.0
```

Would just do, and it would be enough:

```
choria req choria_util machine_state name=check_load --reply-filter 'ok() && semver(data("version"), "< 1.0.0")'
```

If there was a way to create dynamically a cli thats essentially just macros around rpc, kv, external commands etc it would allow them to make a local utility that has easy to discover features without that is just a wrapper.

```
$ ./myco --help
usage: ./myco [<flags>] <command> [<args> ...]

A thing that does Foo

Flags:
  --help  Show context-sensitive help (also try --help-long and --help-man).

Commands:
  help [<command>...]
    Show help.

  machines [<flags>] <machine>
    Query Autonomous Agents
```

```
$ ./myco machines --help
usage: ./myco machines [<flags>] <machine>

Query Autonomous Agents

Flags:
  --help         Show context-sensitive help (also try --help-long and --help-man).
  --not=VERSION  Matches instances that are not this version
  --le=VERSION   Matches instances that are older than or same as
  --lt=VERSION   Matches instances that are older than
  --ge=VERSION   Matches instances that are newer than or same as
  --gt=VERSION   Matches instances that are older than
  --eq=VERSION   Matches instances that are same as

Args:
  <machine>  Autonomous Agent to report on
```

```
$ ./myco machines check_bacula_fd --eq 1.0.0
2022/05/09 14:40:35 choria req choria_util machine_state name=check_bacula_fd --filter-replies ok() && semver(data("version"), "= 1.0.0")
Discovering nodes using the choria method .... 29

29 / 29    0s [==============================================================================] 100%

...

Summary of Name:

   check_bacula_fd: 16

Summary of State:

   OK: 16

Summary of Version:

   1.0.0: 16

Finished processing 29 / 29 hosts in 1.304s
```

This could be defined using a YAML file like this:

```yaml
name: Foo Thing
description: A thing that does Foo
version: 0.0.1
author: R.I.Pienaar <rip@devco.net>
commands:
  - name: machines
    description: Query Autonomous Agents
    std_filter: true
    request:
      agent: choria_util
      action: machine_state
      params:
[...]

PR #1666, opened 2022-05-10
(#1665) initial basic working app framework

ripienaar:
This allows applications to be built that wraps RPC and KV
components.

Application definitions are documented using a schema and
are validated on loading, from the definition a CLI app is
built that can have nested sub commands and more.

Application definitions goes in files called x-app.yaml,
a symlink from x to choria would then load x-app.yaml and
construct the CLI when x is invoked.

An additional config file can be read called applications.yaml
that can be accessed via go templates in the definition in a few
places.

The config and definitions can live in ., ~/.config/choria/builder
or /etc/choria/builder.

Paths are using XDG, but its a super primitive implementation
for the moment, will be refined via issue 1624, thus paths
and file names are subject to change

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -12,11 +12,13 @@ import (
 	"path/filepath"
 	"runtime"
 	"runtime/pprof"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
 
 	"github.com/choria-io/go-choria/protocol"
+	"github.com/choria-io/go-choria/providers/appbuilder"
 
 	log "github.com/sirupsen/logrus"
 	"gopkg.in/alecthomas/kingpin.v2"
@@ -52,6 +54,20 @@ func ParseCLI() (err error) {
 	ctx, cancel = context.WithCancel(context.Background())
 
 	go interruptWatcher()
+
+	// If we are not invoked as something something choria, then check
+	// if the app builder has an app configuration matching the name we
+	// are run as, if it does, we invoke it instead of the standard choria
+	// cli tools
+	//
+	// TODO: too janky, need to do a better job here, looking at the name is not enough
+	if !strings.Contains(os.Args[0], "choria") {
+		builder := appbuilder.NewAppBuilder(ctx, filepath.Base(os.Args[0]))
+		if builder.HasDefinition() {
+			builder.RunCommand()
+			os.Exit(0)
+		}
+	}
 
 	bi = &build.Info{}
 
@@ -210,7 +226,7 @@ func forcequit() {
 		}
 	}
 
-	<-time.NewTimer(grace).C
+	<-time.After(grace)
 
 	dumpGoRoutines()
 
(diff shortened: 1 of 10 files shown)


cmd/cmd.go
```diff
@@ -53,6 +55,14 @@ func ParseCLI() (err error) {
 
 	go interruptWatcher()
 
+	// TODO: too janky, need to do a better job here
+	if !strings.Contains(os.Args[0], "choria") {
+		builder := appbuilder.NewAppBuilder(ctx, filepath.Base(os.Args[0]))
+		if builder.HasDefinition() {
+			builder.RunCommand(ctx)
```
smortex, line 62: I think we want to exit when we have run an "app"?
```suggestion
			builder.RunCommand(ctx)
			return
```

ripienaar: Yeah actually fixed that in a uncommitted change here already. 

I am sure there are many horrors here hehe

smortex: A `return` cause more trouble than it is supposed to fix, I could workaround the issue with this:
```diff
diff --git a/cmd/cmd.go b/cmd/cmd.go
index 7ddcdca8..4774787a 100644
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -50,7 +50,7 @@ var (
 	ran        bool
 )
 
-func ParseCLI() (err error) {
+func ParseCLI() (wantsToRun bool, err error) {
 	ctx, cancel = context.WithCancel(context.Background())
 
 	go interruptWatcher()
@@ -60,6 +60,7 @@ func ParseCLI() (err error) {
 		builder := appbuilder.NewAppBuilder(ctx, filepath.Base(os.Args[0]))
 		if builder.HasDefinition() {
 			builder.RunCommand(ctx)
+			return false, nil
 		}
 	}
 
@@ -81,12 +82,12 @@ func ParseCLI() (err error) {
 		if cmd.FullCommand() == cli.command {
 			err = cmd.Configure()
 			if err != nil {
-				return fmt.Errorf("%s failed to configure: %s", cmd.FullCommand(), err)
+				return false, fmt.Errorf("%s failed to configure: %s", cmd.FullCommand(), err)
 			}
 		}
 	}
 
-	return
+	return true, nil
 }
 
 func systemConfigureIfRoot(actAsServer bool) error {
diff --git a/main.go b/main.go
index e2234ffb..05f16a1d 100644
--- a/main.go
+++ b/main.go
@@ -14,16 +14,19 @@ import (
 
 func main() {
 	var err error
+	var wantsToRun bool
 
-	err = cmd.ParseCLI()
+	wantsToRun, err = cmd.ParseCLI()
 	if err != nil {
 		log.Fatalf("Could not configure Choria: %s", err.Error())
 		os.Exit(1)
 	}
 
-	err = cmd.Run()
-	if err != nil {
-		log.Fatalf("Could not run Choria: %s", err.Error())
-		os.Exit(1)
+	if wantsToRun {
+		err = cmd.Run()
+		if err != nil {
+			log.Fatalf("Could not run Choria: %s", err.Error())
+			os.Exit(1)
+		}
 	}
 }
```

The comment in ParseCLI makes me think you want to do some refactoring here, just adding this to be sure it does not slip through :wink:

ripienaar: did something simple - just had this before your suggestion - think its ok, but as you say, I have some itterations left to do here

Issue #1701, opened 2022-05-30
Improve help output


ripienaar:
Help output has become a multi page crazy affair, do same as in appbuilder etc

PR #1702, opened 2022-05-30
(#1701) use kingpin master

ripienaar:


ripienaar:
@vjanelle not sure if this will annoy some of your efforts, this will no doubt break some CLI apps that use some of the libraries here :( but kingpin is unmaintained and this is the only way to get some code. 

Ultimately going to have to work it i suspect :(

vjanelle:
Thanks for the heads up- we should be ok.

Issue #1768, opened 2022-07-15
lifecycle.Event toString representation


vjanelle:
I need string representations of the lifecycle.Event constants for logging purposes.

ripienaar:
https://github.com/choria-io/go-choria/pull/1769

PR #1769, opened 2022-07-15
(@#1768) Generate string representations of Lifecycle.Event

vjanelle:
* Generate String method and values - sadly golang does not have Enums
  so this is going to be annoying to maintain if we add lots of types
* Fix unhandled error warning in PublishEvent

### DIFF
--- a/lifecycle/lifecycle.go
+++ b/lifecycle/lifecycle.go
@@ -30,10 +30,11 @@ import (
 type Type int
 
 const (
-	_ = iota
+	// Unknown is for when a Type string was passed in that doesn't match what was expected
+	Unknown Type = iota - 1
 
 	// Startup is an event components can publish when they start
-	Startup Type = iota
+	Startup
 
 	// Shutdown is an event components can publish when they shutdown
 	Shutdown
@@ -48,11 +49,29 @@ const (
 	Governor
 )
 
+//lint:ignore U1000 #1768 support for external clients
+func (t Type) String() string {
+	switch t {
+	case Startup:
+		return "Startup"
+	case Shutdown:
+		return "Startup"
+	case Provisioned:
+		return "Provisioned"
+	case Alive:
+		return "Alive"
+	case Governor:
+		return "Governor"
+	default:
+		return "Unknown"
+	}
+}
+
 // Format is the event format used for transporting events
 type Format int
 
 const (
-	// UnknownFormat is a unknown format message
+	// UnknownFormat is an unknown format message
 	UnknownFormat Format = iota
 
 	// ChoriaFormat is classical ChoriaFormat lifecycle events in its own package
@@ -61,6 +80,20 @@ const (
 	// CloudEventV1Format is a classical Choria lifecycle event carried within a version 1.0 CloudEvent
 	CloudEventV1Format
 )
+
+//lint:ignore U1000 #1768 support for external clients
+func (f Format) String() string {
+	switch f {
+	case UnknownFormat:
+		return "UnknownFormat"
+	case ChoriaFormat:
+		return "ChoriaFormat"
+	case CloudEventV1Format:
+		return "CloudEventV1Format"
+	default:
+		return "UnknownFormat"
+	}
+}
 
 var eventTypes = make(map[string]Type)
 var eventJSONParsers = make(map[Type]func([]byte) (Event, error))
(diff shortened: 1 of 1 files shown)


lifecycle/lifecycle.go
```diff
@@ -48,6 +48,20 @@ const (
...
+var typeToString = map[Type]string{
+	Startup:     "Startup",
+	Shutdown:    "Shutdown",
+	Provisioned: "Provisioned",
+	Alive:       "Alive",
+	Governor:    "Govenor",
+}
+
+//lint:ignore U1000 #1768 support for external clients
+func (t Type) toString() string {
```
ripienaar, line 61: make `String()` so just printing the type will do the right thing

lifecycle/lifecycle.go
```diff
@@ -62,6 +76,18 @@ const (
...
 
+//lint:ignore U1000 #1768 support for external clients
+var formatToString = map[Format]string{
+	UnknownFormat:      "UnknownFormat",
+	ChoriaFormat:       "ChoriaFormat",
+	CloudEventV1Format: "CloudEventV1Format",
+}
+
+//lint:ignore U1000 #1768 support for external clients
+func (f Format) toString() string {
```
ripienaar, line 87: as above

lifecycle/lifecycle.go
```diff
@@ -221,7 +247,10 @@ func PublishEvent(e Event, conn inter.RawNATSConnector) error {
 		return err
 	}
 
-	conn.PublishRaw(target, j)
+	err = conn.PublishRaw(target, j)
+	if err != nil {
+		return err
```
ripienaar, line 252: Not sure where we are calling this - but generally publishing events is best efforts so I specifically didnt return eror here, but I guess its harmless to do so

vjanelle: ah

lifecycle/lifecycle.go
vjanelle: Also so that it'd actually be available externally >.<

lifecycle/lifecycle.go
```diff
@@ -62,6 +76,18 @@ const (
...
+//lint:ignore U1000 #1768 support for external clients
+var formatToString = map[Format]string{
+	UnknownFormat:      "UnknownFormat",
+	ChoriaFormat:       "ChoriaFormat",
+	CloudEventV1Format: "CloudEventV1Format",
+}
+
+//lint:ignore U1000 #1768 support for external clients
+func (f Format) String() string {
+	return formatToString[f]
```
ripienaar, line 88: Think we need to be a bit more careful here to also handle `""` for example, so maybe:

```
s, ok := formatToString[f]
if !ok { return fomatToString[UnknownFormat] }
return s
```

Format is probably ok but the Type one we should handle similar and perhaps change `_ = iota` to `Unknown Type = iota` then use that.

As is, when the type 0 for some weird reason we'd get an empty string

Sorry for not noticing that earlier

ripienaar: with 0 never being used and not having any name before - i think we can just call 0 unknown.

I always (well I try) skip 0 in these iotas since just doing `x:=X{}` if that X type holds a key of this type it would be 0, so I want real values set on purpose.  But it means we need to handle 0 gracefully in these kinds of tings

lifecycle/lifecycle.go
```diff
@@ -48,11 +49,29 @@ const (
...
+//lint:ignore U1000 #1768 support for external clients
+func (t Type) String() string {
+	s := map[Type]string{
+		Unknown:     "Unknown",
+		Startup:     "Startup",
+		Shutdown:    "Shutdown",
+		Provisioned: "Provisioned",
+		Alive:       "Alive",
+		Governor:    "Governor",
+	}[t]
```
ripienaar, line 61: This was better before when you had the maps external to the function, this way we allocate a bunch of strings every time we call this then dont use them.

A better way - if you want to do in the function like this would be to use a case statement on t and just return the string, then at least it doesnt allocate all the names for all the types  and the map each time

vjanelle: I can't wait for golang to eventually realize people want Enums and do this natively.

ripienaar (approved):
LGTM

Issue #1779, opened 2022-08-03
Leafnodes broken with latest release


smortex:
After updating to the latest release from the previous one, the leafnode I use to reach my infra was unable to connect to the main broker anymore and logged:

```
{"component":"network_broker","level":"debug","msg":"Will not attempt to connect to remote server on \"puppet.vittoria.pro:7422\" (51.79.18.52:7422), leafnodes currently disabled","time":"2022-08-03T10:15:13-10:00"}
{"component":"network_broker","level":"debug","msg":"Error trying to connect as leafnode to remote server \"puppet.vittoria.pro:7422\" (attempt 4): leafnodes disabled","time":"2022-08-03T10:15:13-10:00"}
```

Nothing in the main broker log.

After downgrading the leafnode to the previous version of choria (0.26.0), it still cannot connect but the main broker, but the main broker now logs something when the leafnode try to connect: "remote leafnode has same cluster name":

```
{"component":"network_broker","level":"info","msg":"202.90.71.171:51834 - lid:76 - Leafnode connection created ","time":"2022-08-03T17:06:17-04:00"}
{"component":"network_broker","level":"debug","msg":"202.90.71.171:51834 - lid:76 - Starting TLS leafnode server handshake","time":"2022-08-03T17:06:17-04:00"}
{"account":"choria","component":"authentication","level":"debug","mTLS":true,"msg":"Registering user '' in account 'choria'","remote":"202.90.71.171:51834","stage":"check","subject":"desktop-fln40kq.lan","time":"2022-08-03T17:06:17-04:00"}
{"component":"network_broker","level":"error","msg":"202.90.71.171:51834 - lid:76 - remote leafnode has same cluster name","time":"2022-08-03T17:06:17-04:00"}
{"component":"network_broker","level":"info","msg":"202.90.71.171:51834 - lid:76 - Leafnode connection closed: Cluster Names Identical account: choria","time":"2022-08-03T17:06:17-04:00"}
{"component":"network_broker","level":"error","msg":"202.90.71.171:51834 - lid:76 - remote leafnode has same cluster name","time":"2022-08-03T17:06:17-04:00"}
```

The main broker configuration for reference:

```
logfile = /var/log/choria-broker.log
loglevel = debug
identity = ns565980.vittoria.pro

plugin.choria.srv_domain = vittoria.pro

# Exposes a number of stats end points:
#   /choria            - JSON information about this broker
#   /choria/prometheus - Prometheus.io compatible statistics
[...]

PR #1780, opened 2022-08-04
(#1779) work around several breaking API changes in nats server

ripienaar:


ripienaar:
So @smortex the good news is that I can make it work, the bad news is Nats Server introduced many breaking changes and we now need to do a bigger release than anticipated since we also have to add some required option to the puppet modules.

You willing to run on 0.26.0 until 0.27.0 is out?

PR #1781, opened 2022-08-05
(misc) remove some races found with -race

ripienaar:


ripienaar:
@vjanelle this fixes all the races i found with `go test -race ./...` only really the client one matters, but probably not even.

We dont run tests with race as its too slow but can do now and then by hand or maybe a scheduled weekly run

vjanelle:
@ripienaar that one in the client code probably explains what I was seeing

Issue #1740, opened 2022-06-22
Create a v2 protocol that does not use x509


ripienaar:
Today the protocol is all x509 all the time, identity is extracted from certs and everyone needs a cert.

We recently did work to introduce JWTs and ed25519 keys for authentication part, today for those to be able to make requests a AAA service which will sign using a x509 key is required

So the goal is essentially to allow jwt+ed25519 to make requests without a AAA service.  This will be a v2 protocol, the server/client needs to be able to support both v1 and v2 protocols at the same time so that where needed a cert can be used for auth+sign to facilitate gradual move to new behaviors as inevitably this will require changes to provisioner, replicator and many other components.

We should look at the mechanics for issuing JWTs a bit more now too as this will essentially end up replacing the role of the CA entirely.  Brokers of course will still need x509 certs to support TLS - but given that we do chain of trust verification on the connection with signatures and keys it might be that TLS using tempoary certs would be entirely possible and safe.

v2 protocol need to also get rid of the whole cached cert thing as JWTs are short lived.  This will therefor probably require a full revisit of security providers.  If possible v1 protocol/security should also remove the cert cache as its just annoying.

The goals and design can be found in the [ADR](https://choria-io.github.io/go-choria/adr/001/index.html).

Changes:

 - [x] for v2 move from `choria:request:2` to `io.choria.protocol.v2.request` etc which is more in line with recent schemas etc
 - [x] Remove the concept of a cert cache entirely from v1 to simplify v2 effort https://github.com/choria-io/go-choria/pull/1842
 - [x] also remove cache from the puppet modules (was not needed as modules are all puppet sec that does it on its own)
 - [x] tokens should all be signable using ed25519 https://github.com/choria-io/go-choria/pull/1743
 - [x] create a permission to grant/deny access to choria fleet in client jwts
 - [x] add design doc https://github.com/choria-io/go-choria/blob/main/protocol/v2/ADR.md
 - [x] move to a `[]byte` orientated api where possible https://github.com/choria-io/go-choria/pull/1844
 - [x] remove string orientated security interface items https://github.com/choria-io/go-choria/commit/ec2cf7e1bd46ff5258724777afe1bca65807d8c3
 - [x] create a network packet analyzer, ongoing refinements
 - [x] when v2 protocol and security is use the default collective should be `choria` https://github.com/choria-io/go-choria/pull/1885
 - [x] Extract schema validation from v1 for reuse between protocols https://github.com/choria-io/go-choria/pull/1852
[...]

PR #1842, opened 2022-09-27
(#1740) remove the concept of a cache from security providers

ripienaar:
The cache was mainly used to find all privileged certificates and then verify signatures against all known privileged certificates.

This was a useless effort since the privileged certificate was already attached with the message.

So this remove the cache entirely and remove privileged checks that was using the cache.

We pulled out the validation that was done during cache save into a new security method and call that instead of save to gain the validation benefits.

Follow up work is to write some integration tests to ensure the security remains safe.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -923,6 +923,8 @@ Path to a file holding tags for a Scout entity
 
 Always store new Public Keys to the cache overwriting existing ones
 
+**This setting is deprecated or already unused**
+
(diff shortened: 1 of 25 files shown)


ripienaar:
I simplified this down to have just the cache related changes, will put the simplifaction and move to []byte orientated functions after this one

ripienaar:
I will merge this one based on our slack conversation, if you find anything more out of line we can address in a follow up

Issue #1882, opened 2022-10-31
Remove the old style choria event format


ripienaar:
We have been using cloud events for a long time, time to retire our old format

PR #1883, opened 2022-10-31
(#1882) Remove the old style event format

ripienaar:


ripienaar:
@vjanelle let me know if you still use the old event format, otherwise I think it can go

ripienaar:
Actually this isnt possible, the cloud event holds a old style event.

Issue #1898, opened 2022-11-09
Support signing signature requests to remote signers


ripienaar:
Relates to https://github.com/choria-io/aaasvc/pull/115

PR #1899, opened 2022-11-09
(#1898) Sign requests to remote signers using the private key

ripienaar:


providers/signers/aaasvc_rpc.go
```diff
@@ -34,7 +35,12 @@ func (s *aaaServiceRPC) Sign(ctx context.Context, request []byte, cfg inter.Requ
 		return nil, err
 	}
 
-	res, err := signer.OptionWorkers(1).Sign(string(request), string(token)).Do(ctx)
+	sig, err := cfg.SignBytes(request)
```
ripienaar, line 38: This should rather ask for the remote signer seed file and sign here using that, this way we can support this for both old and new security systems.

PR #1951, opened 2023-01-04
(#1740) allow unmodified nats clients to do pub-sub in issuer mode

ripienaar:
Often companies use the choria broker for standard nats things as well as choria, choria auth though presents a problem since it is not nats auth.

So rather than demand 100% of connections need to be updated this allow, when running in issuer mode, mTLS connections to access the broker and streams for pub-sub purposes unmodified.  However they do not have access to any choria subjects, cannot observe fleet management traffic or interact with them in any way.

They have full streams and other access though.

In traditional setups we would give such clients full access to all subjects, so this is the analogue to that but with some restrictions.

This is only enabled if mTLS is configured in issuer mode and only if the connection was fully verified using mTLS.

With no CA set or somehow mTLS not passing the connection is denied


ploubser (commented):
👍

ploubser (approved):

PR #1957, opened 2023-01-12
(#1740) Extract the tokens package into a new project

ripienaar:
The tokens package is shared by a number of things like Stream Replicator and others.

Those packages though have no interest in the full dependency tree that depending on go-choria would bring and it creates nasty chicken and egg situations.

We've extracted the tokens package into it's own repository and this starts a series of updates to reference that instead.


ploubser (approved):
lgtm

Issue #1959, opened 2023-01-12
autonomous agent choria built in check reports invalid cert expiry


ripienaar:
When configured in the #1740 org mode the choria health check seems to be confused bout token expiry:

```
choria-compose-server.choria-10      | time="2023-01-12T14:29:04Z" level=info msg="423b9958fdf1.choria.local check_choria#check OK: OK: /var/log/choria-status.json|uptime=1200;; filtered_msgs=0;; invalid_msgs=0;; passed_msgs=1;; replies_msgs=1;; total_msgs=1;; ttlexpired_msgs=0;; last_msg=1673533659;; cert_expire_seconds=-9223372036;;" component=machine
```

It's saying cert expired -9xxxxxx ago which is wrong but also doesnt report token expiry.

It should report the case of having no cert (or token) to check as math.MaxInt expiry.

PR #1960, opened 2023-01-12
(#1959) fix handling of token and cert expires being 0 time

ripienaar:
Also report token expires


ripienaar:
@ploubser see issue details, I spotted this in the choria compose container logs when they check themselves

ploubser (approved):
👍

PR #1961, opened 2023-01-17
(#1740) adjust proto v2 tests to be based on an issuer

ripienaar:
This also expands the tests and does a test verifying that the AAA flow works both happy path and failure paths


ploubser (approved):
lgtm!

Issue #1968, opened 2023-01-20
archivewatcher is a bit aggressive on extremely broken machines


ripienaar:
When a machine is completely broken - like under very high CPU - the governor messages might context timeout, but they might make it to the broker.

In that case the watcher will be a bit aggressive about retrying governor entry such that there could be a gradual build up in the governor.  A few of these broken machines can cause big troubles.

We should mitigate this:

 * Slower governor tries on context timeout
 * Fewer governor tries per run cycle
 * Longer - like 30 minutes even - max backoffs for nodes that are just broken for a very long time

This should flatten things out even in the face of many very bad nodes

PR #1970, opened 2023-01-23
(#1968) only retry governor on governor full states

ripienaar:
To avoid a situation where machine that is failing to process ok messages keep sending more and more gov election messages we now treat anything except a governor full as unretryable

The effect is that in watchers should there be a bad situation with the machine a watcher is on this will now try and fail fast then relying on the watcher intervals for retries which would typically align with governor bucket configuration

### DIFF
--- a/providers/governor/streams/governor.go
+++ b/providers/governor/streams/governor.go
@@ -38,6 +38,7 @@ package governor
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sort"
 	"strings"
@@ -105,6 +106,8 @@ type Manager interface {
 	// Connection is the NATS connection used to communicate
 	Connection() *nats.Conn
 }
+
+var errRetry = errors.New("retryable error")
 
 type jsGMgr struct {
 	name     string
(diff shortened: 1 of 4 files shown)


providers/governor/streams/governor.go
```diff
@@ -288,10 +291,13 @@ func (g *jsGMgr) Start(ctx context.Context, name string) (Finisher, uint64, erro
 
 		res, err := jsm.ParsePubAck(m)
 		if err != nil {
-			if !jsm.IsNatsError(err, 10077) {
-				g.Errorf("Invalid pub ack: %s", err)
+			// jetstream sent us a puback error, this is retryable in the case of governors
+			if jsm.IsNatsError(err, 10077) {
```
ripienaar, line 295: gvernor full is the healthy fail - if succesfully processed we know we can retry safely (its working, just telling us to try again) vs not working which is a fail.

providers/governor/streams/governor.go
```diff
@@ -332,9 +338,19 @@ func (g *jsGMgr) Start(ctx context.Context, name string) (Finisher, uint64, erro
 
 	g.Debugf("Starting to campaign every %v for a slot on %s using %s", g.cint, g.name, g.subj)
 
+	// we try to enter the governor and if it fails in a way thats safe to retry
+	// we will do so else we exit.
+	//
+	// We need to handle thins like context timeout, bucket not found etc specifically
```
ploubser, line 344: `thins` typo

ploubser (commented):
Minor typo but otherwise LGTM

ploubser (approved):

Issue #1976, opened 2023-02-01
machine run should treat connection failure as a soft failure


ripienaar:
The intention for `choria machine run` is to run a auto agent locally without needing working choria.  Unfortunately when KV was added the requirement for a connection came in - but thats not needed for most cases so somehow this hsould be optional.

Now it just fails on connection

PR #1977, opened 2023-02-02
(#1976) Do not connect or initialize config by default

ripienaar:
This enables "choria machine run" to be used in more cases locally without requiring a full working choria setup

Watchers like KV that require network connections will fail with errors logged. To test those a network connection is unavoidable.

### DIFF
--- a/aagent/watchers/gossipwatcher/gossip.go
+++ b/aagent/watchers/gossipwatcher/gossip.go
@@ -163,6 +163,11 @@ func (w *Watcher) startGossip() {
 		}
 
 		publish := func() {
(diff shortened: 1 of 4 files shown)


aagent/watchers/gossipwatcher/gossip.go
```diff
@@ -163,6 +163,11 @@ func (w *Watcher) startGossip() {
 		}
 
 		publish := func() {
+			if !w.ShouldWatch() {
```
ripienaar, line 166: small bug fix these were firing when they shouldnt

ploubser (approved):
lgtm

Issue #1978, opened 2023-02-13
Add --governor to choria jwt server


ripienaar:
Server JWTs support a governor permission but we don't support actually setting that on the CLI

PR #1979, opened 2023-02-13
(#1978) Add --governor to choria jwt server

ploubser:
Here we add --governor to the choria jwt server cli. Check if --stream-user is also set otherwise we exit with an error.


ripienaar:
Looks good, but wtf is up with tests let me see maybe need to update the actions

ripienaar (approved):
LGTM, merge when ready

PR #1983, opened 2023-02-15
(#1982) Tag AAA Login urls with comma split

ripienaar:
Ensure they will correctly load balance and randomize


patch
```diff
@@ -1,289 +0,0 @@
-diff --git a/broker/network/choria_auth.go b/broker/network/choria_auth.go
```
ripienaar, line 1: doh, removed this left over file that shouldnt be here.

ploubser (approved):

PR #1986, opened 2023-02-17
(#1987) Export SetBuildBasedOnJWT

ploubser:
resolver.setBuildBasedOnJWT is a method that other proftarget plugins often would like to use.

Here we decouple the SetBuildBasedOnJWT method from the resolver struct and export it as a function. To make the function usable outside of the resolver struct we add the target jwt file, build.info and reader function parameters.

Tests have also been updated to use the new function signature.

### DIFF
--- a/providers/provtarget/builddefaults/default.go
+++ b/providers/provtarget/builddefaults/default.go
@@ -7,6 +7,7 @@ package builddefaults
 import (
 	"context"
 	"fmt"
+	"io"
 	"net"
 	"os"
 	"strings"
@@ -55,7 +56,14 @@ func (b *Resolver) Configure(cfg *config.Config, log *logrus.Entry) {
 
 	b.identity = cfg.Identity
 
-	_, err := b.setBuildBasedOnJWT()
+	jwtreader, err := os.Open(b.bi.ProvisionJWTFile())
+	if err != nil {
+		log.Errorf("Configuration of the provisioner settings based on JWT file %s failed: %s", jwtf, err)
+		return
+	}
+	defer jwtreader.Close()
+
+	_, err = SetBuildBasedOnJWT(jwtreader, b.bi)
 	if err != nil {
 		log.Errorf("Configuration of the provisioner settings based on JWT file %s failed: %s", jwtf, err)
 	}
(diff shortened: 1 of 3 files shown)


providers/provtarget/builddefaults/default.go
```diff
@@ -113,11 +113,9 @@ func (b *Resolver) Targets(ctx context.Context, log *logrus.Entry) []string {
 	return servers.Strings()
 }
 
-// setBuildBasedOnJWT sets build settings based on contents of a JWT file
-func (b *Resolver) setBuildBasedOnJWT() (*tokens.ProvisioningClaims, error) {
-	bi := b.bi
-
-	d, err := os.ReadFile(bi.ProvisionJWTFile())
+// SetBuildBasedOnJWT sets build settings based on contents of a JWT file
+func SetBuildBasedOnJWT(jwtf string, bi *build.Info, readFilef func(string) ([]byte, error)) (*tokens.ProvisioningClaims, error) {
```
ripienaar, line 117: Hmm, think its a bit non idiomatic with the readFilef here.

Typically you'd do like `SetBuildBasedOnJWT(bi *build.Info, jwt string)` I get that it would probably help a bit in callers to make their code shorter and more generic but would rather do like this.

An alternative would be something like this

```go
func SetBuildBasedOnJWTFile(filesys fs.FS,path string){
	f,_:=filesys.Open(path)
	jwt,_:=io.ReadAll(f)

	// do stuff
}
```

You then do:

```
SetBuildBasedOnJWTFile(os.DirFS("/tmp"), "x.jwt")
```

Of if you have an embedded fs you do `SetBuildBasedOnJWTFile(internal.FS, "x.jwt")` or similar, I quite like this.

ripienaar: Will need an issue opened to help me with changelogs - you can also update chnagelog at the same time if you want (i always forget so no biggie)

ploubser: Cool, I'll take a swing at that.

providers/provtarget/builddefaults/default.go
```diff
@@ -113,16 +121,16 @@ func (b *Resolver) Targets(ctx context.Context, log *logrus.Entry) []string {
...
-	bi := b.bi
-
-	d, err := os.ReadFile(bi.ProvisionJWTFile())
-	if err != nil {
-		return nil, err
+// SetBuildBasedOnJWT sets build settings based on contents of a JWT file
+func SetBuildBasedOnJWT(r io.Reader, bi *build.Info) (*tokens.ProvisioningClaims, error) {
+	jwt := make([]byte, 1024)
+	for {
+		_, err := r.Read(jwt)
```
ripienaar, line 128: ```jwt,err=io.ReadAll(r)``` iirc

ripienaar (changes requested):
small change else lgtm

ripienaar (approved):
LGTM

PR #1993, opened 2023-02-21
(#1740) allow v1 provisioners, without a token, to connect to a issuer brokers

ripienaar:
It's possible we need to support a fleet with mix old and new servers wishing to be provisioned. The ones that are too old for v2 network might be provisioned onto another network, or defered after recording hostnames or similar to create hit lists for upgrades.

This allows, after opt-in, provisioners with just user+pass+tls to access the provisioner account.  Those provisioners can then choose to provision old nodes differently.


broker/network/choria_auth.go
```diff
@@ -508,6 +477,24 @@ func (a *ChoriaAuth) handleProvisioningUserConnectionWithTLS(c server.ClientAuth
...
+
+	opts := c.GetOpts()
+
+	if opts.Password == emptyString {
+		return false, fmt.Errorf("password required")
+	}
+
+	if a.provPass != opts.Password {
+		return false, fmt.Errorf("invalid provisioner password supplied")
+	}
```
ripienaar, line 496: some monitor refactoring here to pull out shared code between the 2 handler functions below

ploubser (approved):
LGTM

Issue #1998, opened 2023-03-07
Support leader election for tally


ripienaar:
Tally should support clustered deployments by using leader election, we should add an `active` label to all metrics and dashboard should select just active metrics.

This avoids the scenario where n tally instances all report the same data and the resulting dashboard essentially has n* the data. With this update the dashboard will now just pick one of the tally instances and the tally instances will on their own elect a leader who will essentially own the stats

When no election is set active is always true.

PR #1999, opened 2023-03-07
(#1998) Support leader election for Tally

ripienaar:
Events will now have a new label added and the value will flip between 0 and 1 depending on leader state


ploubser (approved):
LGTM

PR #2000, opened 2023-03-08
(#1740) improve deny rules for clients with no permissions

ripienaar:
For clients with no permissions set today we have
a open publish ruleset with no allow or deny, this ensures we deny all when that is the intention.


ploubser (approved):
👍

Issue #2010, opened 2023-03-21
Surface builtin type as plugin for the nagios watcher


ripienaar:
At present there is no way to know when looking at events if its from goss/heartbeat/nagios_status - the builtin plugins - so we should set the plugin to the builtin value in events for those.

PR #2011, opened 2023-03-21
(#2010) Record the builtin type as plugin in nagios watchers

ripienaar:


aagent/watchers/nagioswatcher/nagios.go
```diff
@@ -446,6 +446,8 @@ func (w *Watcher) watchUsingPlugin(ctx context.Context) (state State, output str
 }
 
 func (w *Watcher) watchUsingBuiltin(_ context.Context) (state State, output string, err error) {
+	w.previousPlugin = w.properties.Builtin
```
ripienaar, line 449: only functional change here

ploubser (approved):

Issue #2014, opened 2023-03-28
Allow dynamic agent loading and unloading


ripienaar:
We wish to support loading agents at run-time, for example for the `external` provider, should an agent be updated we need to replace the already loaded agent with a new stub.  Or if a agent is removed we should be able to remove interest in those messages.

This is to support puppet-free deployments where we will devise a method to deploy agents into running servers similar to what is supported for autonomous agents using the [Machines Manager](https://github.com/choria-io/go-choria/tree/main/aagent/watchers/machineswatcher).

For now lets see how we can facilitate this in AgentManager and Connector and later look into the agent providers.

PR #2015, opened 2023-03-28
(#2014) Support run-time replacing and removing agents

ripienaar:


ripienaar:
This is just the core capability in connections and agent manager.  Nothings using it yet.

choria/connection.go
```diff
@@ -77,7 +78,6 @@ type Connection struct {
 	log               *log.Entry
 	fw                *Framework
 	config            *config.Config
-	subscriptions     map[string]*nats.Subscription
```
ripienaar, line 80: unused code removed

PR #2018, opened 2023-03-29
(misc) update dependencies

ripienaar:
Also fix ginkgo version missmatch


ripienaar:
Looks like github actions is having a holiday.

ploubser (approved):

Issue #2019, opened 2023-03-29
Allow external agents to be placed entirely in a sub directory


ripienaar:
To facilitate machines manager like deployment of external agents it would be ncie to place them in a directory holding all files - ddls and all.

that way to swap in a new agent we can prepare new in tmp, move current to tmp, move new from tmp, nuke current out of tmp

While doing this, external agent should also ignore the directory `tmp` under the libdir just like machines does

PR #2020, opened 2023-03-29
(#2019) allow external agents to live in directories

ripienaar:
This adjusts external agents to live entirely in a sub directory that can be moved in an atomic way to later ease some deployment scenarios.


ploubser (approved):
👍

PR #2021, opened 2023-03-30
(#2019) support relocating agents at runtime

ripienaar:
Also fixes a panic in util when checking path related errors.

Also adjusts when recently modified ddls are excxluded so that they are not considered orphaned and removed before being added again on next reconcile


ploubser (approved):
LGTM

Issue #2022, opened 2023-03-31
Implement a native sha256


ripienaar:
Today we call out to the unix sha256 command to verify archives:

https://github.com/choria-io/go-choria/blob/main/aagent/watchers/archivewatcher/archive.go#L331-L346

But this is not portable and just gross, we should implement a compatible sha256 internal to choria and support generating those files

PR #2023, opened 2023-03-31
(#2022) create a native sha256 checker

ripienaar:
Use it in the archive watcher and enable archive watcher in standard builds


ploubser (approved):
LGTM

Issue #2029, opened 2023-04-10
Support deploying external agents with the machines watcher


ripienaar:
Later we'll rename it to plugins watcher once its properly generic.

jeffmccune:
You're probably aware, but just an FYI the `choria machine mms` command isn't plumbed to match the [docs](https://choria.io/docs/autoagents/watcher_reference/#autonomous-agent-archives) yet.  I took a stab but wasn't sure if this should be part of the new plugin system.  Only mention it in off chance it was overlooked.

jeffmccune:
Never-mind, I see it: `choria machine plugins pack`

ripienaar:
Yeah sorry about that been a bit of turbulence here as I realised it would be impossible to get the “machines” watcher to be more generic and keep working as before. 

So there is a plugins watcher now that’s almost identical but can also handle rpc agents and more. 

Docs not updated for that will try today. 

Such is life on main I am afraid

PR #2033, opened 2023-04-11
(#2029) Create plugins watcher

ripienaar:
This makes the machines watcher more generic and renames it to plugins watcher, the standard build is updated to include this instead of machines watcher.

For backward compatability the old machines watcher is restored in the code base for later deprecation.

Command line was updated to use new watcher


ploubser (approved):
Awesome 👍

PR #2034, opened 2023-04-11
(#2029) Extract some common code and reuse in CLI

ripienaar:
Also improve the plugins pack cli.

The external agents provider will check frequently at first for new agents then settle into a 30s kind of interval, with jitter.  This optimise bootup to deliver agents quickly but with a more relaxed settled state


ploubser (approved):
Minor typo found by the linter but otherwise lgtm

Issue #2037, opened 2023-04-13
Handle trailing newline when KV watcher sets machine data


jeffmccune:
Problem:

The KV watcher stores machine data as a string instead of a JSON object when the watched value contains a trailing newline.  This is a problem for the plugins watcher which then fails with `Managing plugins failed: '' expected a map, got 'string'" component=machine`

Expected Behavior:

The KV watcher should ignore trailing white space when parsing the value as json for use as nested machine data.  A trailing newline is likely, for example:

```sh
choria machine plugins pack plugins.json plugins.seed | choria kv put PLUGINS machines -
```

Work around:

```
choria machine plugins pack plugins.json plugins.seed | tr -d '\n' | choria kv put PLUGINS machines -
```

jeffmccune:
I'm working on a fix for this.

PR #2038, opened 2023-04-13
(#2037) Trim space when parsing kv machine data

jeffmccune:
Problem:
The KV watcher stores the value from the bucket into machine data as a string instead of a JSON object when the value contains a trailing newline.  When this happens, the plugins watcher fails with `Managing plugins failed: '' expected a map, got 'string'" component=machine`

Solution:
Ignore leading and trailing white space from the watched value when the kv watcher checks the data for `{...}` or `[...]`.

Result:
The following command, which adds a trailing newline, is deserialized into machine data as a map value instead of a string value.  The plugins watcher works successfully.

```
choria machine plugins pack plugins.json plugins.seed | choria kv put PLUGINS machines -
```

Resolves: #2037


jeffmccune:
@ripienaar Tossed this draft up so you could take a peek if I'm on the right path with the mocking.  I followed the examples in the existing agents.

From a design perspective, I plan to create a slice or something to find the start and end non-whitespace characters.  I get the sense you looked at the start and end to avoid the cost of trying to parse the whole byte array blindly.

ripienaar (approved):
Looks great, thank you.  Let me know if you're 100% on this PR and I can merge

jeffmccune:
Yes I’m 100% done with it. Thanks for taking a look.

ripienaar:
@jeffmccune mind signing your commit?

jeffmccune:
Don’t mind, one min.

jeffmccune:
Signature added.

Issue #2057, opened 2023-04-20
Adapters using in-process connections log SlowConsumer errors


ripienaar:
These are harmless and everything works but quite annoying to see

PR #2058, opened 2023-04-20
(#2057) Avoid calling the nats in process provider to test if it exists

ripienaar:
We would call InProcessConn() and then check if its nil to determine if the feature was enabled then later we would call again to create the connection.

Problem is this first call actually made a client in the server but we never attached a client to it on the other side so eventually it would be detected as dead and discarded.  The real connection would continue to work as normal.

This would log a bunch of slow consumer errors etc that was hard to debug, we fix this now by checking if the provider was set rather than calling it


ploubser (approved):
LGTM

Issue #2059, opened 2023-04-24
Add ways to verify packed machines/plugins specifications


ripienaar:
These should be usable in CI to verify correct data is sent to the fleet

PR #2060, opened 2023-04-24
(#2059) Support verifying plugins and machines data

ripienaar:


ploubser (approved):
👍

ploubser:
Couple of test failures but looks sound otherwise. Can merge on green.

PR #2075, opened 2023-08-09
(#2076) Add plugin.nats.connect_timeout / MCOLLECTIVE_NATS_CONNECT_TIMEOUT option

johnhcaldwell:
Added to support choria servers connecting to a broker cluster which takes a long time to establish a connection, greater than the NATS default of 2 seconds.  Accepts a duration as a parameter, but will apply the default value if none is specified.

### DIFF
--- a/choria/connection.go
+++ b/choria/connection.go
@@ -751,6 +751,9 @@ func (conn *Connection) Connect(ctx context.Context) (err error) {
 		options = append(options, nats.UserCredentials(conn.config.Choria.NatsCredentials))
 	}
 
+	conn.log.Warnf("Setting NATS connect timeout to %v", conn.config.Choria.NatsConnectTimeout)
+	options = append(options, nats.Timeout(conn.config.Choria.NatsConnectTimeout))
+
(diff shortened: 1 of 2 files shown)


choria/connection.go
```diff
@@ -751,6 +751,9 @@ func (conn *Connection) Connect(ctx context.Context) (err error) {
 		options = append(options, nats.UserCredentials(conn.config.Choria.NatsCredentials))
 	}
 
+	conn.log.Warnf("Setting NATS connect timeout to %v", conn.config.Choria.NatsConnectTimeout)
```
ripienaar, line 754: Probably can be info here.

ripienaar (changes requested):
Looks good thanks, minor comment there.

Please review the contributors guide and file an issue and update commits as detailed there.

ripienaar:
Have to run go fmt on go code before comitting - config file needs update.

Please squash commits to one and also update the commit to the same format `(#xxx) ...`

ripienaar:
closing stale PR, feel free to re-open and resolve fmt issues

PR #2082, opened 2023-09-08
(#2077) Support setting audit log group and mode

### LINKED ISSUE #2077: Choria server audit logs should have configurable permissions
The Choria server audit logs should be able to be configured to something like mode 0640 and also set the group ownership to someone other than root.  This will allow tools like Vector or Splunkforwarder to collect the logs since those applications tend to run as unprivileged users.

treydock:
Allows tools like Splunk Forwarder or Vector to access the audit log without elevated privileges such as running as root
Fixes #2077

### DIFF
--- a/providers/agent/mcorpc/audit/choria.go
+++ b/providers/agent/mcorpc/audit/choria.go
@@ -10,7 +10,10 @@ package audit
 import (
 	"encoding/json"
 	"fmt"
+	"io/fs"
 	"os"
+	"os/user"
+	"strconv"
 	"sync"
 	"time"
 
@@ -41,9 +44,17 @@ func Request(request protocol.Request, agent string, action string, data json.Ra
 	}
 
 	logfile := cfg.Option("plugin.rpcaudit.logfile", "")
+	logfileGroup := cfg.Option("plugin.rpcaudit.logfile.group", "")
+	logfileModeOpt := cfg.Option("plugin.rpcaudit.logfile.mode", "0600")
 
 	if logfile == "" {
 		log.Warnf("Choria RPC Auditing is enabled but no logfile is configured, skipping")
+		return false
+	}
+
+	logfileMode, err := strconv.ParseUint(logfileModeOpt, 0, 32)
+	if err != nil {
+		log.Errorf("Failed to parse plugin.rpcaudit.logfile.mode: %s", err)
 		return false
 	}
 
@@ -67,7 +78,7 @@ func Request(request protocol.Request, agent string, action string, data json.Ra
 	mu.Lock()
 	defer mu.Unlock()
 
-	f, err := os.OpenFile(logfile, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0600)
+	f, err := createAuditLog(logfile, logfileGroup, logfileMode)
 	if err != nil {
 		log.Warnf("Auditing is not functional because opening the logfile '%s' failed: %s", logfile, err)
 		return false
@@ -81,4 +92,35 @@ func Request(request protocol.Request, agent string, action string, data json.Ra
 	}
 
 	return true
+}
+
+func createAuditLog(logfile string, logfileGroup string, logfileMode uint64) (*os.File, error) {
+	f, err := os.OpenFile(logfile, os.O_CREATE|os.O_APPEND|os.O_WRONLY, fs.FileMode(logfileMode))
+	if err != nil {
+		return f, err
+	}
+
+	if logfileGroup == "" {
+		return f, nil
+	}
+
+	grp, err := user.LookupGroup(logfileGroup)
+	if err != nil {
+		f.Close()
+		return f, err
+	}
+
+	gid, err := strconv.Atoi(grp.Gid)
+	if err != nil {
+		f.Close()
+		return f, err
+	}
+
+	err = os.Chown(logfile, os.Getuid(), gid)
+	if err != nil {
+		f.Close()
+		return f, err
+	}
+
+	return f, nil
 }
--- a/providers/agent/mcorpc/audit/choria_test.go
+++ b/providers/agent/mcorpc/audit/choria_test.go
@@ -6,6 +6,7 @@ package audit
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"runtime"
 	"testing"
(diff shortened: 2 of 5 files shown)


treydock:
Marked draft as I'd like to see if I can add some tests.  Wanted to get feedback on the implementation though.  Not thrilled about the deep conditional to handle the group lookup and chown but wasn't sure how else to handle all the possible errors that could arise.

providers/agent/mcorpc/audit/choria.go
```diff
@@ -41,12 +44,20 @@ func Request(request protocol.Request, agent string, action string, data json.Ra
 	}
 
 	logfile := cfg.Option("plugin.rpcaudit.logfile", "")
+	logfileGroup := cfg.Option("plugin.rpcaudit.logfile.group", "")
+	logfileModeOpt := cfg.Option("plugin.rpcaudit.logfile.mode", "0600")
```
ripienaar, line 48: Probably worth adding all these 3 to teh config classes as options so that they are discoverable using things like `choria tool config` and the generated docs. 

I can deal with that in a follow up as looks like there might be some annoyance with getting that in.  So just a comment here for now that I'll do that later.

providers/agent/mcorpc/audit/choria.go
```diff
@@ -67,13 +78,30 @@ func Request(request protocol.Request, agent string, action string, data json.Ra
...
+		if err == nil {
+			gid, err := strconv.Atoi(grp.Gid)
+			if err == nil {
+				err = os.Chown(logfile, os.Getuid(), gid)
+				if err != nil {
+					log.Warnf("Unable to chown file %s to group %s: %s", logfile, logfileGroup, err)
+				}
+			} else {
+				log.Warnf("Unable to convert group %s GID: %s", logfileGroup, err)
+			}
```
ripienaar, line 99: You cn void these nested ifs by making a function that creates the file, then you can just return early and so avoid all the else stuff.

This is actually quite difficult to do right, since in this code your file will exist for a short while as the wrong user - open as root then chown - this means your file tailers could fail for a while and depending on how tjhey handle that scenario could prevent the file from being accessed.

Safer to make a tmp file in the same directory (has to be the same directory) chown the temp file then rename it and pass back the file handle

This means opening the file becomes:

 1. Open for append and set the file handle
 2. If that fails, call into the create function that returns the new file handle
 3. Log to the file handle

This avoids the race and the file only ever exist in the intended name with the right permissions and ownership

ripienaar (changes requested):
Looks good, but handling files as different users are unfortunately more involved :(

providers/agent/mcorpc/audit/choria.go
treydock: Both Splunk Forwarder and Vector can handle brief periods of file inaccessibility as they will retry if the file is created and momentarily the wrong GID.  This is almost certainly a non issue for most file tailers since none can guarantee a file ever exists so they almost always will retry to access a file since things like OS boot it's not guaranteed a file will exist yet so the tail programs are almost always designed to retry.

Is it worth the extra logic to handle something that could very well be a non-issue?  Since the audit log won't even exist until a request comes through, any tailer being used with Choria would have to be able to handle the absence of the file and retry and absence and lack of read permission will generally look the same to programs going to open the file.

ripienaar: No you're right thats ok, lets try to remove the nested ifs though

providers/agent/mcorpc/audit/choria_test.go
```diff
@@ -1,13 +1,19 @@
 // Copyright (c) 2020-2021, R.I. Pienaar and the Choria Project contributors
 //
 // SPDX-License-Identifier: Apache-2.0
+//go:build !windows
+// +build !windows
```
treydock, line 5: I couldn't find another way to avoid Windows tests failing because using `syscall.Stat_t` to validate with unit tests, since seems the code still has to compile even when wrapped in `runtime.GOOS != "windows"`.

treydock:
@ripienaar I updated to use function to get rid of the nested conditionals.  I tried to add some unit tests but they will fail on Windows due to using `syscall.Stat_t`.  Wasn't sure if I should just remove the tests entirely or what the best way forward is.

providers/agent/mcorpc/audit/choria_test.go
ripienaar: make a file `choria_test_windows.go` inside it 

```
func checkFileGid(nfo *os.FileInfo) { }
```

and in `choria_test_unix.go`

```
func checkFileGid(nfo *os.FileInfo) {
  // do the check 
}
```

and from this file just call `checkFileGid(nfo)` it will then compile fine everywhere just noop on windows

ripienaar (changes requested):
Great improvement, I think the mode checks on windows is useful so added some hints on maybe how to fix the test

providers/agent/mcorpc/audit/choria_test.go
treydock: Implemented this suggestion

providers/agent/mcorpc/audit/choria.go
```diff
@@ -41,12 +44,20 @@ func Request(request protocol.Request, agent string, action string, data json.Ra
...
+	logfileModeOpt := cfg.Option("plugin.rpcaudit.logfile.mode", "0600")
 
 	if logfile == "" {
 		log.Warnf("Choria RPC Auditing is enabled but no logfile is configured, skipping")
 		return false
 	}
 
+	logfileMode, err := strconv.ParseUint(logfileModeOpt, 0, 32)
+	if err != nil {
+		log.Warnf("Failed to parse plugin.rpcaudit.logfile.mode: %s", err)
```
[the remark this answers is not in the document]
treydock: Updating this log message

providers/agent/mcorpc/audit/choria_test_unix.go
```diff
@@ -0,0 +1,27 @@
...
+
+package audit
+
+import (
+	"os"
+	"os/user"
+	"strconv"
+	"syscall"
+
+	"github.com/onsi/gomega"
```
treydock, line 15: Couldn't use dot import due to staticcheck:

```
providers/agent/mcorpc/audit/choria_test_unix.go:15:2: should not use dot imports (ST1001)
```

treydock:
@ripienaar I've made the necessary adjustments to testing and also the logging severity suggestion.

providers/agent/mcorpc/audit/choria_test_unix.go
ripienaar: Thats pretty weird, I'll see if I can fix it post merge, probably staticheck doesnt figure this is a test file

ripienaar:
Looks great, squash and mark is not draft :)

treydock:
@ripienaar Commits are squashed

ripienaar (approved):
LGTM

PR #2083, opened 2023-09-20
(#2077) Some fixes to rpc audit

ripienaar:
* We use declared options in the config structs to show up in 'choria tool config'
 * We write to a temp file for the audit log to avoid races between tests
 * We only try to set the group when ran as root
 * We use the nobody group which is in more systems by default
 * Fix some file naming shenanigans to make vet happy


providers/agent/mcorpc/audit/choria_unix_test.go
```diff
@@ -12,16 +12,16 @@ import (
 	"strconv"
 	"syscall"
 
-	"github.com/onsi/gomega"
+	. "github.com/onsi/gomega"
```
treydock, line 15: For my own curiosity, how did you get staticcheck to not dislike this dot import?

treydock:
@ripienaar Looks good.  Only question is just curiosity as I ran into lint issues when I tried to do the dot import.  Using `nobody` is much better as that will work on OS X which is where I was doing the development and means I don't have to do uncommitted hacks to use local groups for testing.

providers/agent/mcorpc/audit/choria_unix_test.go
ripienaar: I believe staticcheck only cares about non test files for this check.  A file called `x_test_unix.go` is not a test but `x_unit_test.go` is a test 

Thats my theory anyway - renaming files is first thing I tried

Issue #2087, opened 2023-10-18
"choria broker server check consumer" command fails with "consumer load failure: context deadline exceeded"


ploubser:
When running the new check consumer command, `choria broker server check consumer` the check fails with output `consumer load failure: context deadline exceeded`. 

Update the `prepareNatsCli` function in cmd/broker.co to include checking the command string for `broker server check` before setting `cfg.Choria.NatsUser` and `cfg.Choria.NatsPass`

PR #2088, opened 2023-10-18
(#2087) Exclude `broker server check consumer` command when using system user

ploubser:
Here we add `broker server check consumer` to the list of cli sub commands we exclude when setting the nats user to the system user.


ripienaar (approved):
LGTM

Issue #2112, opened 2024-02-08
CHORIA_SRV_DOMAIN not getting used from environment variable


treydock:
Now that Choria 0.28.0 is released trying to validate #2097

```
$ export CHORIA_SRV_DOMAIN=dev.company.net
$ choria tool config plugin.choria.srv_domain
<SNIP>
Configuration item: plugin.choria.srv_domain

║        Value: <GCP project domain, not dev.company.net>
║          URL: https://choria.io/docs/deployment/dns/
║    Data Type: string
║  Environment: CHORIA_SRV_DOMAIN
║
║ The domain to use for SRV records, defaults to the domain the
║ server FQDN is in
╙─
```

It looks like the lack of a default is preventing this code path from getting used.  Will come up with a simple fix.

PR #2113, opened 2024-02-08
(#2112) Ensure CHORIA_SRV_DOMAIN is actually getting used

treydock:
The logic to pull config defaults from an environment variable is only used when a config has a default. This adds empty string default for SRV domain which will trigger the logic to pull a default value also from the CHORIA_SRV_DOMAIN environment variable.


treydock:
Got a local build working on Mac and did this:

```
$ export CHORIA_SRV_DOMAIN=dev.company.net
$ ./choria-0.28.0-darwin-amd64 tool config
<SNIP>
Connectivity:

       Main Collective: mcollective
           Collectives: mcollective
      Service Registry: disabled
            SRV Domain: dev.company.net

```

ripienaar:
Are you 100% sure you updated? For me this works fine:

```
% export CHORIA_SRV_DOMAIN=dev.company.net
% choria tool config srv
.....
            SRV Domain: dev.company.net
.....
Configuration item: plugin.choria.srv_domain

║        Value: dev.company.net
║          URL: https://choria.io/docs/deployment/dns/
║    Data Type: string
║  Environment: CHORIA_SRV_DOMAIN
║
║ The domain to use for SRV records, defaults to the domain the
║ server FQDN is in
╙─

```

ripienaar:
There's definitely something pretty weird going on, let me check, this should be working but testing in a docker container for example it does not

treydock:
@ripienaar Any update if this could get merged? We've been using a custom patched RPM and would like to not have to depend on custom built RPMs to get this change.

ripienaar:
The main problem here is I dont believe the problem is that a default is not set, as I mentioned this works outside of docker without this change.

From the test we have a field here without a default:

https://github.com/choria-io/go-choria/blob/1d4f659d077ef171f59937e581dda195625e24e9/confkey/confkey_test.go#L26

And testing it here, it works:

https://github.com/choria-io/go-choria/blob/1d4f659d077ef171f59937e581dda195625e24e9/confkey/confkey_test.go#L290-L302

So there is something else going on, we should rather look to fix the real problem.

So asking again, are you doing this in Docker or a normal machine?

treydock:
> So asking again, are you doing this in Docker or a normal machine?

These are running inside a Docker container that runs the Choria client code. 

The issue https://github.com/choria-io/go-choria/issues/2112 shows it not working using a regular VM, not Docker.  The environment we consume this patch is Docker but the issue exists in both places.

FWIW this patch solved the issue but I haven't dug into any other possible solutions.

ripienaar:
Hmm, not sure why previously for me there was a working case for a VM but not a container, this is indeed a behaviour where environments are only parsed for keys with defaults.

It kind of make sense because if you go down the path to set values it just doesn't know the difference between unset and set and ends up inserting empty strings in weird places which surfaced a lot of other weird things when I tried to fix that.

So adding a default seems best

ripienaar:
Thanks @treydock

PR #2170, opened 2024-05-02
(#2169) Pass choria federations to external discovery agent

vjanelle:


config/config.go
```diff
@@ -255,6 +255,7 @@ func NewConfigForTests() *Config {
 	c.DisableSecurityProviderVerify = true
 	c.LogFile = "discard"
 	c.RPCAuthorization = false
+	c.Choria.FederationCollectives = []string{"alpha", "beta"}
```
vjanelle, line 258: Couldn't figure out how to get things to work via os.Setenv :\

inter/imocks/util.go
```diff
@@ -96,6 +96,7 @@ func NewFrameworkForTests(ctrl *gomock.Controller, logWriter io.Writer, opts ...
 	fw.EXPECT().Configuration().Return(mopts.cfg).AnyTimes()
 	fw.EXPECT().Logger(gomock.AssignableToTypeOf("")).Return(logrus.NewEntry(logger)).AnyTimes()
 	fw.EXPECT().NewRequestID().Return(util.RandomHexString(), nil).AnyTimes()
+	fw.EXPECT().FederationCollectives().Return([]string{}).AnyTimes()
```
vjanelle, line 99: This is probably incorrect now

providers/discovery/external/external.go
```diff
@@ -40,12 +40,13 @@ type Response struct {
...
-	Protocol   string            `json:"protocol"`
-	Collective string            `json:"collective"`
-	Filter     *protocol.Filter  `json:"filter"`
-	Options    map[string]string `json:"options"`
-	Schema     string            `json:"$schema"`
-	Timeout    float64           `json:"timeout"`
+	Protocol    string            `json:"protocol"`
+	Collective  string            `json:"collective"`
+	Filter      *protocol.Filter  `json:"filter"`
+	Federations []string          `json:"federations"`
```
vjanelle, line 46: This will require updating the schema, unsure if this is what you meant or if it should be something in filter

config/config.go
ripienaar: go has some issues getting environment in containers, i dont know we had another case of it not working reported and I couldnt figure it out.

providers/discovery/external/external.go
ripienaar: no this is good, exactly what I had in mind.  We already pass Collective there so it seemed a related good fit.

providers/discovery/external/external_test.go
```diff
@@ -95,3 +93,7 @@ var _ = Describe("External", func() {
 		})
 	})
 })
+
+func RunExternalCommands() {
```
ripienaar, line 97: this maybe is some left over?

vjanelle: yup, forgot to push this up.

config/config.go
vjanelle: Do you want me testing for the case where federation is not setup?

vjanelle: Also what was weird was that I could do os.Setenv(), and then do os.Getenv() and retrieve the value, but fetching it out from fw.FederationCollectives wouldn't do the right thing.  I'll take a longer look at it.

vjanelle: Figured it out, it does work.  Need to setup the mocks though.

providers/discovery/external/options.go
```diff
@@ -35,6 +36,13 @@ func Collective(c string) DiscoverOption {
 	}
 }
 
+// Federations returns the list of federated collectives to discover in
```
ripienaar, line 39: ...sets the federated collectives to discover in

ripienaar (commented):
looks pretty good, minor edit

providers/discovery/external/options.go
vjanelle: oops

cmd/cmd.go
```diff
@@ -194,6 +196,11 @@ func commonConfigure() error {
 		return fmt.Errorf("could not parse configuration: %s", err)
 	}
 
+	// If list of federations is specified on the CLI, mutate the configuration directly
+	if len(federations) > 0 {
```
vjanelle, line 200: @ripienaar what I thought I might have to do to support overriding the federations at the cli, otherwise I can create a method in Choria to set the federations

cmd/discover.go
```diff
@@ -44,6 +51,14 @@ func (d *discoverCommand) Setup() error {
 
 func (d *discoverCommand) Configure() error {
 	err = commonConfigure()
+
+	// If list of federations is specified on the CLI, mutate the configuration directly
+	if len(d.federations) > 0 {
+		cfg.Choria.FederationCollectives = strings.Split(d.federations, ",")
+	} else {
+		fmt.Println("We are not setting federations")
```
ripienaar, line 59: some left over debug here

cmd/req.go
```diff
@@ -462,6 +466,13 @@ func (r *reqCommand) Configure() error {
 		return err
 	}
 
+	// If list of federations is specified on the CLI, mutate the configuration directly
+	if len(r.federations) > 0 {
+		cfg.Choria.FederationCollectives = strings.Split(r.federations, ",")
+	} else {
+		fmt.Println("We are not setting federations")
```
ripienaar, line 473: some more debugging left over here

ripienaar (changes requested):
small changes otherwise its good

vjanelle:
Added the flag to the `ping` command as well, I use that for making sure connectivity is ok

ripienaar (approved):
LGTM

Issue #1105, opened 2021-01-23
dont parse plugin.d configs if under home


ripienaar:
Till now we'd parse ~/.mcollective then look for ~/plugin.d/*.cfg for stuff, this never made sense. The plugin.d is pretty much only in /etc.

So we'll just not even try to parse those if the active config is in ~

PR #2173, opened 2024-05-17
Fix config load when HOME="/"

robert-fraczek:
There is an issue i found on FreeBSD systems with starting choria-server via "service choria-server start" command causing global configuration doesn`t load. Same problem occur after OS reboot. Moreover i discover that starting via /usr/local/etc/rc.d/choria-server work properly.
Problem is with env HOME set to "/" when starting with service command ( rc.d set HOME="/root" ).

There is an issue with logic discovering local "home" configuration that prevent global configs to load.
Matching $HOME as prefix to config path when $HOME is "/" will always be true preventing to load global configs - as a quick fix i propose to make an exception for that.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -410,7 +410,7 @@ func (c *Config) UnParsedOptions() map[string]string {
 func (c *Config) dotdDir() string {
 	if !forceDotParse {
 		home, err := iu.HomeDir()
-		if err == nil {
+		if err == nil && home != "/" {
 			if strings.HasPrefix(c.ConfigFile, home) {
 				return ""
 			}


config/config.go
```diff
@@ -410,7 +410,7 @@ func (c *Config) UnParsedOptions() map[string]string {
 func (c *Config) dotdDir() string {
 	if !forceDotParse {
 		home, err := iu.HomeDir()
-		if err == nil {
+		if err == nil && home != "/" {
```
ripienaar, line 413: This seems like quite a big special case here that just doesnt seem like it belong.  Should we fix FreeBSD rc scripts instead @smortex ?

smortex:
@robert-fraczek, I am not sure to follow, and I can't reproduce your problem.

Are you saying that the home directory of the root user is the root of the filesystem, i.e. `/` instead of the usual `/root`?

Forcing `HOME` to `/` does not seems to cause any trouble as far as I can see:
```sh-session
# HOME=/ choria server run --config /usr/local/etc/choria/server.conf
```

Maybe you can describe the whole configuration files you have because my setup is probably quite different than yours?

smortex:
Thank you for the details, I think I better understand what is going on, and as you say, I think it is the combination of multiple issues that break your use case.

I have no issue with my configuration where I only rely on `plugin.choria.srv_domain` in the system-wide configuration file `/usr/local/etc/choria/server.conf`.  The fact that the server capacity of loading your `plugin.d/choria.cfg` depends on the $HOME environment variable feels wrong, especially when the path in this variable does not appear in the path to the configuration file.

### About `HOME=/` when using `service(8)` on FreeBSD

This is documented in the `service(8)` man page:
> When used to run rc.d scripts the service command sets HOME to /

It was introduced in:
https://github.com/freebsd/freebsd-src/commit/ee55fdb8fa1ce60d39bd7f2afde7d9772a2e3d58

It is unlikely to change, and I would not expect a system-wide service like the choria server to rely on the value of `$HOME` for anything.  Changing the behavior of the `choria` rc.d script is of course possible, but that does not feel the right fix: the program should rather work regardless of the value of the `HOME` environment variable.

### How `HOME=/` affects choria

When run as a client (`choria req ...`), choria can load additional configuration files form the current working directory and its parent directories up to the root of the file system.  All the loaded configuration files are listed in the output of `choria tool config`, and indeed this list unexpectedly change when `HOME` is set to something that match the beginning of the path of the "User Config" file:

```sh-session
romain@zappy ~/Opus-Codium % echo $HOME
/home/romain/Opus-Codium
romain@zappy ~/Opus-Codium % choria tool config | awk '/Configuration Files/,/Security Configuration/'
Configuration Files: 

           User Config: /usr/local/etc/choria/client.conf
          Loaded Files: /usr/local/etc/choria/client.conf
                        /usr/local/etc/choria/plugin.d/actionpolicy.cfg
                        /usr/local/etc/choria/plugin.d/choria.cfg
                        /usr/local/etc/choria/plugin.d/puppet.cfg
                        /home/romain/choria.conf
                        /home/romain/Opus-Codium/choria.conf

Security Configuration: 
romain@zappy ~/Opus-Codium % HOME=/ choria tool config | awk '/Configuration Files/,/Security Configuration/'
Configuration Files: 

           User Config: /usr/local/etc/choria/client.conf
          Loaded Files: /usr/local/etc/choria/client.conf
                        /home/romain/choria.conf
                        /home/romain/Opus-Codium/choria.conf

Security Configuration: 
romain@zappy ~ % HOME=/usr choria tool config | awk '/Configuration Files/,/Security Configuration/'
Configuration Files: 

           User Config: /usr/local/etc/choria/client.conf
          Loaded Files: /usr/local/etc/choria/client.conf
                        /home/romain/choria.conf
                        /home/romain/Opus-Codium/choria.conf

Security Configuration: 
romain@zappy ~/Opus-Codium % 
```

I recall we did not want to load `plugins.d/*` config from all the user directories when loading extra configuration files, but the code you change predate that and is from #1105.

Giving it a try, if I write a file `~/.choriarc`, this one is loaded *instead* of the system-wide `client.conf` file, but `~/plugin.d/*` config files are not loaded.   Because the system-wide config is not loaded either, that hardly make sense to me.

If we remove the whole conditional block, the code become independent of the value of `$HOME`, and one can totally ignore system-wide config with his own config files in `~/.choriarc` and `~/plugins.d`.  Do you know if this would cause some kind of issue @ripienaar?

ripienaar:
The problem is we cant totally ignore /etc/.../plugins.d as many of the module plugins set their settings there and some are even application settings.  Thus if we skip them we'd skip important things. This led to undefined behaviour simply by trying to override some settings in your home dir. 

Essentially think of /etc/.../plugins.d being puppet managed settings that should be enabled for the system to work.  But puppet wouldnt manage them anywhere else so the current behaviour kind of achieves that.

The behaviour of the choria.conf you added @smortex is much better and more useful, but we're kind of stuck with this mess around the other files.

smortex:
I struggled with go to try to implement what I understand to be the correct behavior in #2174: determine the system-wide location for choria config, and use this to load plugins regardless of the configuration being used.

PR #2174, opened 2024-05-22
(#2173) Load plugins configuration from the system directory

smortex:
Some plugins need to load configuration that is managed system-wide by Puppet even when the user is using their own configuration file. Regardless of the configuration file loaded (system one or user one), we want to load these plugins configuration from the system.

We previously inferred the plugins configuration directory relatively to the rest of the configuration, but ignored it if it was located in the user HOME directory.  This was not correct, and even caused more trouble when the HOME environment variable was '/' as it is commonly on some systems for services, because plugins configuration was never loaded on these systems.

Fix this by detecting the location of the system configuration and unconditionally loading plugins configuration from this directory.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -96,6 +96,9 @@ type Config struct {
 	// ConfigFile is the main configuration that got parsed
 	ConfigFile string
 
+	// The system-wide configuration directory. Plugins are loaded from there
+	SystemConfigDirectory string
+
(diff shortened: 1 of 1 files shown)


smortex:
Oh god, fixing the CI looks so :facepalm:.  Before changing a trillion lines of code to introduce some kind of structure to allow passing a custom directory, I would like to be sure that this is going in the right direction. Thanks!

PR #2203, opened 2024-10-22
container images typically don't have systemd, this should allow the choria package to be installed on those images.

TomRitserveldt:


smortex (commented):
Well, I trend to think that container images should not have more that a binary and its required libs (so no shell, no apt, etc)… :smile: 

In any case, don't mix tabs and spaces (stick to tabs to match what is used in these files), and make sure the last line has its trailing '\n' (it is probably wise to configure your editor to do it for you).

ripienaar:
Please review the contributing guide and follow that, we need issues filed and specific forms commit messages.

ripienaar:
@TomRitserveldt you planning to update this or shall I close it?

thanks

vjanelle:
@ripienaar could potentially use this in the future so created a issue for them.

Issue #2076, opened 2023-08-09
Add support for configurable NATS connection timeout


johnhcaldwell:
For choria servers operating under certain conditions, the default NATS connection timeout (2 seconds) is insufficient.  A config file / environment variable for setting the `nats.Timeout()` option prior to establishing connections is needed.

ripienaar:
Thank you

PR #2259, opened 2025-07-14
(#2076) Add configurable authentication timeout for NATS brokers

vjanelle:


config/docstrings.go
```diff
@@ -97,6 +98,8 @@ var docStrings = map[string]string{
 	"plugin.choria.registration.file_content.compression":          "Enables gzip compression of registration data",
 	"plugin.choria.registration.inventory_content.compression":     "Enables gzip compression of registration data",
 	"plugin.choria.registration.inventory_content.target":          "NATS Subject to publish registration data to",
+	"plugin.choria.registration.size_trigger":                      "Enables a trigger that will publish a registration message if the size of the message has changed by greater than the trigger amount in bytes",
```
vjanelle, line 101: I re-ran the doc generator and I guess this was missing

config/choria.go
```diff
@@ -81,6 +81,7 @@ type ChoriaPluginConfig struct {
 	NetworkSystemPassword              string        `confkey:"plugin.choria.network.system.password"`                                                             // Password used to access the Choria system account
 	NetworkSystemUsername              string        `confkey:"plugin.choria.network.system.user"`                                                                 // Username used to access the Choria system account
 	NetworkTLSTimeout                  int           `confkey:"plugin.choria.network.tls_timeout" default:"2"`                                                     // Time to allow for TLS connections to establish, increase on slow or very large networks
+	NetworkAuthTimeout                 int           `confkey:"plugin.choria.network.auth_timeout" default:"2"`                                                    // Time to allow for clients to authenticate, increase on slow or very large networks
```
ripienaar, line 84: I think better to make this a duration and then when setting it call `.Seconds()` on it, gives people some more flexibility and its a bit more consistent with things like write deadline below

ripienaar: Should maybe also error if this is over some silly value like 1 mnute

broker/network/network.go
```diff
@@ -55,13 +55,19 @@ func NewServer(c inter.Framework, bi BuildInfoProvider, debug bool) (s *Server,
 		mu:      &sync.Mutex{},
 	}
 
+	// Validate timeouts
+	if s.config.Choria.NetworkAuthTimeout > 60*time.Second || s.config.Choria.NetworkTLSTimeout > 60*time.Second {
```
vjanelle, line 59: @ripienaar I have a version of this that uses the validator package instead, but when I looked in there I couldn't help but refactor it so that might to be a different PR

ripienaar: Yeah validator is a bit long in the tooth for sure, be nice to get things to it for expressing limits like this.

broker/network/network.go
```diff
@@ -55,13 +55,19 @@ func NewServer(c inter.Framework, bi BuildInfoProvider, debug bool) (s *Server,
...
+		return nil, fmt.Errorf("NetworkAuthTimeout and NetworkTLSTimeout must not exceed 60 seconds")
```
ripienaar, line 60: Elsewhere like here https://github.com/choria-io/go-choria/blob/a8957975c7dc9f232d41d92eb5f1eea4d4d55243/broker/network/network.go#L161 we use the config name rather than structure name to help people figure out what it is thats wrong - they dont know the structure names

ripienaar (changes requested):
Looks good except the one thing about errors!

vjanelle:
@ripienaar done - I'll stage a PR about the validator changes - I assume we can use generics now?

ripienaar:
We probably could yes though the syntax is like being stabbed in the eye. Also slow.

ripienaar (approved):

Issue #2265, opened 2025-08-20
Missing packages for Debian 13 (Trixie)


smortex:
Debian 13 was recently released.  For the puppet-choria [repo URI](https://github.com/choria-io/puppet-choria/blob/0b639b84d12b2e2eccf1b1cfb03ed9a65c68bcab/manifests/repo.pp#L77) to work as expected, we must publish packages [here](http://mirrorlists.choria.io/apt/release/debian/trixie).

ripienaar:
I also put it here https://apt.de.choria.io/misc/

smortex:
> Cool! I will have a look next week at $WORK

Installed flawlessly, works as expected.  We can ship it!

Malte-Wagner:
Hi @ripienaar, may you release it at https://apt.de.choria.io/release/debian/trixie ?

Malte-Wagner:
Sounds reasonable.

PR #2266, opened 2025-08-20
(#2265) Build packages for Debian 13 (trixie)

smortex:
Debian 13 was recently released.  Adjust our tooling to build packages
for this new version of Debian.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -5,6 +5,8 @@ on:
       - '*'
 
 jobs:
+  # RedHat
+
   el8_64:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -125,6 +129,48 @@ jobs:
...
           version: tag
 
+  trixie_64:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Build
+        uses: choria-io/actions/packager@main
+        with:
+          build_package: trixie_64
+          packager_tag: trixie-go1.24
```
ripienaar, line 139: first need to make the packager stuff in choria-io/packager

ripienaar: Are any of these Debian releases now unsupported? Can we remove some?

PR #2282, opened 2025-10-15
(#2281) Add a configurable connect timeout for NATS

### LINKED ISSUE #2281: Add a configurable connect timeout for NATS
Choria has configurable timeouts for Brokers, such as `plugin.choria.network.tls_timeout` and `plugin.choria.network.auth_timeout`, but does not provide a way to configure the connect timeout for servers connecting to a broker. NATS uses a 2-second timeout for connections by default: https://github.com/nats-io/nats.go/blob/main/nats.go#L58. Increasing the timeouts on the broker might not be enough in a slow network or an overloaded broker (e.g., during a reconnection storm). Thus, to properly set the timeouts between server and broker, I'd like a configuration option on Choria to set the connect timeout on the server side.

andersonvaf:
Closes #2281


ripienaar:
We should probably just make the client/servers configure itself using the same settings as the brokers as those are named generically `plugin.choria.network.tls_timeout` rather than add more settings

ripienaar (commented):
Auth and TLS timeout are different things. 

TLS connections - depending on algo and version - can take a few packets to establish, this has a separate timeout.  Auth timeout is how long the server waits for auth from the client and time for some internal processing.  In Choria Broker auth timeout is not that big a deal though if you have big TLS certs and using TLS auth this part can also take a fair while. So we have both these settings.

So in retrospect I think I was wrong, holiday mode brain!, to say use auth/tls timeouts here since the go client just have `nats.Timeout()` we should probably rather add `plugin.choria.network.connect_timeout` and ensure that is clamped to below `120s` like the broker does for auth+tls, when adding that add a comment there indicating it affects only clients 

Sorry for sending you on the wrong direction!

andersonvaf:
I changed the code to add the connect_timeout and updated the descriptions. I also ran the doc generator. Let me know if I need to undo the docs.

I was incorrect regarding the behavior of TLS and auth timeouts. The auth timer is scheduled only after the TLS handshake (https://github.com/nats-io/nats-server/blob/main/server/server.go#L3445C1-L3450C3). And by doing so, TLS and Auth timeouts aggregate to a final timeout, indeed. Thanks for pointing that out.

PR #2347, opened 2026-05-17
(misc) Debian: Add  option to prefix choria-server process

bastelfreak:
For EL systems, we already have an option to prefix the process with another binary, for example ionice. This option was missing on Debian. To stay consistent, I added it for Debian as well.

### DIFF
--- a/packager/templates/debian/generic/server.service
+++ b/packager/templates/debian/generic/server.service
@@ -5,7 +5,7 @@ After=network.target
 [Service]
 User=root
 Group=root
-ExecStart={{cpkg_bindir}}/{{cpkg_name}} server --config={{cpkg_etcdir}}/server.conf
+ExecStart=/bin/sh -c "${COMMAND_PREFIX} {{cpkg_bindir}}/{{cpkg_name}} server --config={{cpkg_etcdir}}/server.conf"
 KillMode=process
 
 [Install]


packager/templates/debian/generic/server.service
```diff
@@ -5,7 +5,7 @@ After=network.target
 [Service]
 User=root
 Group=root
-ExecStart={{cpkg_bindir}}/{{cpkg_name}} server --config={{cpkg_etcdir}}/server.conf
+ExecStart=/bin/sh -c "${COMMAND_PREFIX} {{cpkg_bindir}}/{{cpkg_name}} server --config={{cpkg_etcdir}}/server.conf"
```
ripienaar, line 8: For el we have this file, should probably add similar ?

https://github.com/choria-io/go-choria/blob/main/packager/templates/el/global/server.sysconfig

ripienaar:
Did you have a use case for this? I'd also consider removing it from RedHat if you didn't have a use case

bastelfreak:
I don't have a usecase for this and it always looked odd to me. I'm fine with removing it.

ripienaar:
Yeah lets do that, I think the use case I had at one specific user went away also

bastelfreak:
Done https://github.com/choria-io/go-choria/pull/2351

PR #2356, opened 2026-08-01
(#2362) feat: Add openvox 9 support

bastelfreak:
openvox 9 removed the `--configprint` option. it was deprecated in the source code for ages. `puppet config print` is the successor and available in *every* openvox agent. So this is an enhancment by adding openvox 9 support, it's not a breaking change.

### DIFF
--- a/puppet/puppetwrapper.go
+++ b/puppet/puppetwrapper.go
@@ -107,9 +107,9 @@ func (p *Wrapper) AIOCmd(command string, def string) string {
 	return path
 }
 
-// Setting retrieves a config setting by shelling out to puppet apply --configprint
+// Setting retrieves a config setting by shelling out to puppet config print
 func (p *Wrapper) Setting(setting string) (string, error) {
-	args := []string{"apply", "--environment", "production", "--configprint", setting}
+	args := []string{"config", "print", "--section", "user", setting}
 
 	out, err := exec.Command(p.AIOCmd("puppet", "puppet"), args...).Output()
 	if err != nil {


bastelfreak:
I tested `puppet config print ssldir` and of course it works. However I didn't recompile choria.

puppet/puppetwrapper.go
```diff
@@ -107,9 +107,9 @@ func (p *Wrapper) AIOCmd(command string, def string) string {
 	return path
 }
 
-// Setting retrieves a config setting by shelling out to puppet apply --configprint
+// Setting retrieves a config setting by shelling out to puppet config print
 func (p *Wrapper) Setting(setting string) (string, error) {
-	args := []string{"apply", "--environment", "production", "--configprint", setting}
+	args := []string{"config", "print", setting}
```
ripienaar: hmm, previously we'd only check under `[apply]` (or main of course merged) now what's the default for `config print`?

bastelfreak: I updated it to use the apply section. If a value isn't set it there, the value from main will be used.

ripienaar:
Plot thickens :)

```
Error: Could not parse /etc/puppetlabs/puppet/puppet.conf: Illegal section 'apply' in config file at (file: /etc/puppetlabs/puppet/puppet.conf, line: 11). The only valid puppet.conf sections are: [main, server, master, agent, user]. Please use the directory environments feature to specify environments. (See https://puppet.com/docs/puppet/latest/environments_about.html)
```

So probably the right thing is to set it to `user` as you can't have apply in a section in puppet.conf

If you could open a issue with this that'd be great

bastelfreak:
@ripienaar I created the github issue. Can you have another look please?

smortex (approved):
Tested successfully on my test infrastructure.

